PR #3, opened 2023-11-11
Update strings.xml

Patrick44fr:
Hey, my little contribution for french language.

thank you for your good work


philipabbey:
I did wonder if this would happen with language support. We have a rather blunt instrument for the translation, using a Python script to automatically translate English to the other languages. This means that any manual changes get overwritten each time we execute it. As such we can't accept improvements to the translation files until we have a solution to not overwriting the manually provided amendments. This is something we have talked about, e.g. by having a tag in the XML file that says "do not touch", and not deleting the XML files before we read the tags.

I'll put this on hold for now. I need @JosephAbbey to take a look. Then we'll be able to pull in your change.

JosephAbbey (commented):
Looks good.

JosephAbbey:
We need a solution to not overwrite corrections for the script.

JosephAbbey:
This will be accepted when https://github.com/house-of-abbey/GarminHomeAssistant/issues/5 is closed

JosephAbbey (approved):

JosephAbbey (approved):

PR #6, opened 2023-11-13
Updated translate script to use corrected strings

JosephAbbey:
```xml
<string corrected="true" id="blah">Blah Blah</string>
```

Will not be updated

### DIFF
--- a/HomeAssistant.code-workspace
+++ b/HomeAssistant.code-workspace
@@ -2,10 +2,6 @@
 	"folders": [
 		{
 			"path": "."
-		},
-		{
-			"name": "abc",
-			"path": "../blah/abc"
 		}
 	],
 	"settings": {}
--- a/translate.py
+++ b/translate.py
@@ -89,7 +89,7 @@ with open("./resources/strings/strings.xml", "r") as f:
   for l in languages:
     os.makedirs(f"./resources-{l[0]}/strings/", exist_ok=True)
     try:
-      with open(f"./resources-{l[0]}/strings/strings.xml", "r") as r:
+      with open(f"./resources-{l[0]}/strings/corrections.xml", "r") as r:
         curr = BeautifulSoup(r.read().replace('\r', ''), features="xml")
     except FileNotFoundError:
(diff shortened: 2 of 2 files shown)


philipabbey:
https://developer.garmin.com/connect-iq/core-topics/resources/ Just wanted to check what you thought of using the translatable=false attribute? It would be more in keeping with the existing schema.

JosephAbbey:
That attribute specifies whether the string should be translated by ConnectIQ so if we have `translatable="true"`, it may try to translate them itself. This attribute also will prevent translation per string not per language so it will not work for corrections.

philipabbey:
A quick test as follows:

```xml
<strings>
    <string id="AppName">HomeAssistant</string>
    <string id="MenuItemOn" corrected="true">On</string>
```

```
ERROR: venu: A:\Philip\Programming\MonkeyC\HomeAssistant\resources\strings\strings.xml:17: Unable to validate resource file: cvc-complex-type.3.2.2: Attribute 'corrected' is not allowed to appear in element 'string'.
ERROR: A string resource matching the provided app name can't be found. You must provide a valid string resource for the app name.
```

So we can't use custom attributes in the XML unless we amend the validation...

JosephAbbey:
We will need to find another way to embed this data. Second XML file?

JosephAbbey:
```xml
<?xml version="1.0" encoding="utf-8"?>
<!--

   Distributed under MIT Licence
     See https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/LICENSE.


   GarminHomeAssistant is a Garmin IQ application written in Monkey C and routinely
   tested on a Venu 2 device. The source code is provided at:
              https://github.com/house-of-abbey/GarminHomeAssistant.

   P A Abbey & J D Abbey, 14 November 2023

-->

<!--
  Corrections for the French language
  Corrections pour la langue française
-->

<corrections>
  <string id="MenuItemOn">Allumée</string>
</corrections>
```

philipabbey (approved):

PR #7, opened 2023-11-14
Changes required for 2-tap-or-toggle-dont-work-on-venu-2

philipabbey:
Made the 'service' field for taps mandatory.
This makes scripts work for both the emulator and real devices. Provide a response for automations started with a tap, and now they can be used for both tap and toggle.

### DIFF
--- a/source/HomeAssistantMenuItem.mc
+++ b/source/HomeAssistantMenuItem.mc
@@ -77,35 +77,34 @@ class HomeAssistantMenuItem extends WatchUi.MenuItem {
             }
             WatchUi.pushView(new ErrorView(strApiUrlNotFound), new ErrorDelegate(), WatchUi.SLIDE_UP);
         } else if (responseCode == 200) {
-            var d = data as Lang.Array;
+            if (Globals.scDebug) {
+                System.println("HomeAssistantMenuItem onReturnExecScript(): Service executed.");
+            }
+            var d     = data as Lang.Array;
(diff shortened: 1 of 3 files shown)


source/HomeAssistantMenuItem.mc
```diff
@@ -77,36 +77,35 @@ class HomeAssistantMenuItem extends WatchUi.MenuItem {
...
+                WatchUi.showToast(toast, null);
+            }
+            if (Attention has :vibrate) {
+                Attention.vibrate([
+                    new Attention.VibeProfile(50, 100), // On  for 100ms
+                    new Attention.VibeProfile( 0, 100), // Off for 100ms
+                    new Attention.VibeProfile(50, 100)  // On  for 100ms
+                ]);
+            }
+            if (!(WatchUi has :showToast) && !(Attention has :vibrate)) {
```
JosephAbbey, line 100: If we cannot `showToast` but we can `vibrate` then there will be no visual indicator.

JosephAbbey (changes requested):

source/HomeAssistantMenuItem.mc
philipabbey: Yes, that's what I intended to code. If you get at least one type of confirmation then you don't need the fallback option. Its just something to acknowledge the tap.

JosephAbbey (approved):

PR #15, opened 2023-11-19
Icons for Menu Types

Someone0nEarth:


philipabbey:
Hi Someone,

We're liking what you've done. The `HomeAssistantMenuItemFactory` now becomes more of a necessity, and we understand the factorisation of the API calls into a `HomeAssistantService` since the two menu item types have a different inheritance structure. The service class also provides future possibilities for refactoring the code in toggle menu items too in order to coalesce API code, but you leave that for another increment another day!

We were wondering if we could extend your `HomeAssistantService` idea? We were thinking:

1. We could economise on each menu item having their own service, by each menu item sharing a single service.

2. The service calls would then need to be made with `mIdentifier` as a parameter. I believe `mIdentifier` is already a member of both `WatchUi.MenuItem` and `WatchUi.IconMenuItem` and hence available.

```monkey
function callService() as Void {
    mHomeAssistantService.call(mIdentifier);
}
```

3. Your new factory class could insert the central service in to each menu item it manufactures. The factory class could own the service.

Thank you for rebasing the changes from main. I appreciate that would have caused some code churn for you.

Philip & Joseph

source/HomeAssistantService.mc
```diff
@@ -0,0 +1,172 @@
...
+    ) {
+        strNoPhone          = WatchUi.loadResource($.Rez.Strings.NoPhone);
+        strNoInternet       = WatchUi.loadResource($.Rez.Strings.NoInternet);
+        strNoResponse       = WatchUi.loadResource($.Rez.Strings.NoResponse);
+        strApiFlood         = WatchUi.loadResource($.Rez.Strings.ApiFlood);
+        strApiUrlNotFound   = WatchUi.loadResource($.Rez.Strings.ApiUrlNotFound);
+        strUnhandledHttpErr = WatchUi.loadResource($.Rez.Strings.UnhandledHttpErr);
+        mApiKey             = Properties.getValue("api_key");
+        mService = service;
+        mIdentifier = identifier;
```
JosephAbbey, line 49: These should be passed as parameters to the call function, rather than stored in the instance

JosephAbbey (changes requested):

philipabbey:
I've got this on to my watch to play with.



Worth noting how the text of long lines wraps. An annoyance with the SDK here is that wrapped lines are centred when unwrapped ones are left justified. Nothing we can do about that unless we adopt the `CustomMenuItem` approach, but I would would prefer to keep things simple.

**Please feel free to make the "Lean UI" the new default.**

That leaves us wondering about the benefit of retaining the old UI as an option. Presently the reason is because we might want to extend the use of sub text to cover situations where the the on/off status does not apply so directly to the entity, e.g. covered/uncovered or in/out. Just checking the API, it looks like we can add sub titles to `IconMenuItem`s, so that indicates we can adopt the lean UI as the only option in future, which would bring simplification.

philipabbey:
The SDK is screwy for justification, not dissimilar to the conversation we had earlier. No option for centred is provided by `WatchUi.MenuItem.MENU_ITEM_LABEL_ALIGN_*`, even though it is the default.

Perhaps `WatchUi.MenuItem.MENU_ITEM_LABEL_ALIGN_RIGHT` it is really intended for languages written right-to-left? Hmm, perhaps we need to base the alignment on whether the device's language is Arabic or Hebrew? One for another issue.

Someone0nEarth:
Hi Philip and Jospeh,

thanks for your review!

> We're liking what you've done. The `HomeAssistantMenuItemFactory` now becomes more of a necessity, and we understand the factorisation of the API calls into a `HomeAssistantService` since the two menu item types have a different inheritance structure. The service class also provides future possibilities for refactoring the code in toggle menu items too in order to coalesce API code, but feel free to leave that for another increment another day by another person!

Yes, you are right, with introducing the `HomeAssistantService`, it would be a good call to use it for toggle too. Not sure if I will do it in this increment... I will think about it, and than decide.

> We were wondering if we could extend your `HomeAssistantService` idea? We were thinking:
> 
>     1. We could economise on each menu item having their own service, by each menu item sharing a single service.

I like this proposal, I was thinking that too. But I didn`t wanted to change the existing "behavior". 

>     2. The service calls would then need to be made with `mIdentifier` as a parameter. I believe `mIdentifier` is already a member of both `WatchUi.MenuItem` and `WatchUi.IconMenuItem` and hence available.

👍

>     3. Your new factory class could insert the central service in to each menu item it manufactures. The factory class could own the service.

👍

Have fun,

  someone

source/HomeAssistantService.mc
Someone0nEarth: mIdentifier is used within the callback method:

```
} else if (responseCode == 200) {
            if (Globals.scDebug) {
                System.println("HomeAssistantService onReturnCall(): Service executed.");
            }
            var d     = data as Lang.Array;
            var toast = "Executed";
            for(var i = 0; i < d.size(); i++) {
                if ((d[i].get("entity_id") as Lang.String).equals(mIdentifier)) {
                    toast = (d[i].get("attributes") as Lang.Dictionary).get("friendly_name") as Lang.String;
                }
            }
```

Someone0nEarth:
> I've got this on to my watch to play with.

Nice :)

Btw: When "installing" an app with usb (so not using the Garmin app store), I can`t set settings for that app. Do you have a workaround for this?

> Worth noting how the text of long lines wraps. An annoyance with the SDK here is that wrapped lines are centred when unwrapped ones are left justified. Nothing we can do about that unless we adopt the `CustomMenuItem` approach, but I would would prefer to keep things simple.

It is annoying... but I am still on the "prefer to keep things simple" side too.

> **Please feel free to make the "Lean UI" the new default.**
> 
> That leaves us wondering about the benefit of retaining the old UI as an option. Presently the reason is because we might want to extend the use of sub text to cover situations where the the on/off status does not apply so directly to the entity, e.g. covered/uncovered or in/out. Just checking the API, it looks like we can add sub titles to `IconMenuItem`s, so that indicates we can adopt the lean UI as the only option in future, which would bring simplification.

👍

I will change the icon one as the default.

Someone0nEarth:
> The SDK is screwy for justification, not dissimilar to the conversation we had earlier. No option for centred is provided by `WatchUi.MenuItem.MENU_ITEM_LABEL_ALIGN_*`, even though it is the default.

Maybe proposing a "centered" option in the Garmin forum?

> Perhaps `WatchUi.MenuItem.MENU_ITEM_LABEL_ALIGN_RIGHT` it is really intended for languages written right-to-left? Hmm, perhaps we need to base the alignment on whether the device's language is Arabic or Hebrew? One for another issue.

Was thinking that too.

philipabbey:
> Btw: When "installing" an app with usb (so not using the Garmin app store), I can`t set settings for that app. Do you have a workaround for this?

Yes I have a "beta" app on the Garmin Connect App Store (https://apps.garmin.com/en-US/apps/db1e9356-8e04-4d6b-ab43-5c70c49b2488) and I upload an exported `.IQ` file to there, then download to my phone. It allows me to continually test small changes, and gives me the setting menu on my phone. That's why there are two `<iq:application id=` values in the manifest header.

I suppose the alternative is to have some temporary code that [sets the properties](https://developer.garmin.com/connect-iq/api-docs/Toybox/Application/Properties.html#setValue-instance_function) the way you want and continually side load. That's annoying for different reasons.

philipabbey:
> Yes, you are right, with introducing the HomeAssistantService, it would be a good call to use it for toggle too. Not sure if I will do it in this increment... I will think about it, and than decide.

Don't bust a gut. Once we have pulled your code onto `main` we can make smaller incremental changes between us.

source/HomeAssistantService.mc
JosephAbbey: Ah yes, in that case it should be passed through as `context`, see: https://developer.garmin.com/connect-iq/api-docs/Toybox/Communications.html#makeWebRequest-instance_function.

philipabbey: Note: `makeWebRequest` with context needs a different callback function definition. It would be something like:

```monkey
private function onReturnCall(responseCode as Lang.Number, data as Null or Lang.Dictionary or Lang.String, context as Lang.Object) as Void {
  <do something with 'context' parameter>
}
```

Then to send a `Lang.Object` in the context use something like:

```monkey
function call() as Void {
    var options = {
        :method  => Communications.HTTP_REQUEST_METHOD_POST,
        :headers => {
            "Content-Type"  => Communications.REQUEST_CONTENT_TYPE_JSON,
            "Authorization" => "Bearer " + mApiKey
        },
        :responseType => Communications.HTTP_RESPONSE_CONTENT_TYPE_JSON,
        :context => <Some Object>
    };        
    :
            Communications.makeWebRequest(
                url,
                {
                    "entity_id" => id
                },
                options,
                method(:onReturnCall)
            );
    :
}
```

Someone0nEarth:
Hello again,

icon types / lean ui is now the default. I would suggest, that we merge this PR into main. And all the refactorings will be done in another PR.

Have fun,
 someone

philipabbey:
I'll have a go at centralising the service later today and commit to your pull request if that's okay. We'll get that structure sorted and then the smaller increments like migrating code from the toggle menu item can be done later.

Someone0nEarth:
Sure :-) 

I would have time later today too. So, if you were unable to start it later, just ping me if you want, and I could begin with it.

source/HomeAssistantService.mc
philipabbey: Done.

philipabbey:
@Someone0nEarth & @JosephAbbey Requesting review of the latest changes I've committed tonight.

source/HomeAssistantView.mc
```diff
@@ -89,10 +70,17 @@ class HomeAssistantView extends WatchUi.Menu2 {
...
     function getItemsToUpdate() as Lang.Array<HomeAssistantToggleMenuItem> {
         var fullList = [];
-        var lmi = mListMenuItems as Lang.Array<HomeAssistantViewMenuItem>;
-        for(var i = 0; i < lmi.size(); i++) {
-            fullList.addAll(lmi[i].getMenuView().getItemsToUpdate());
+        
+        var lmi = mListMenuItems as Lang.Array<WatchUi.MenuItem>;
+        for(var i = 0; i < mListMenuItems.size(); i++) {
+            var item = lmi[i];
+            if (item instanceof HomeAssistantViewMenuItem) {
```
JosephAbbey, line 77: Duplicated code, why not

```monkey
if (item instanceof HomeAssistantViewMenuItem || item instanceof HomeAssistantViewIconMenuItem) {
  fullList.addAll(item.getMenuView().getItemsToUpdate());
}
```

source/HomeAssistantView.mc
```diff
@@ -124,14 +112,27 @@ class HomeAssistantViewDelegate extends WatchUi.Menu2InputDelegate {
             if (Globals.scDebug) {
                 System.println(haItem.getLabel() + " " + haItem.getId());
             }
-            haItem.execScript();
+            haItem.callService();
```
JosephAbbey, line 115: Why not use an or to remove duplicated code.

JosephAbbey (changes requested):

source/HomeAssistantView.mc
philipabbey: Done.

source/HomeAssistantView.mc
philipabbey: No duplicate code in this region. Types/classes are different.

JosephAbbey:
```monkey
        } else if (item instanceof HomeAssistantMenuItem) {
            var haItem = item as HomeAssistantMenuItem;
            if (Globals.scDebug) {
                System.println(haItem.getLabel() + " " + haItem.getId());
            }
            haItem.callService();
        } else if (item instanceof HomeAssistantIconMenuItem) {
            var haItem = item as HomeAssistantIconMenuItem;
            if (Globals.scDebug) {
                System.println(haItem.getLabel() + " " + haItem.getId());
            }
            haItem.callService();
        } else if (item instanceof HomeAssistantViewMenuItem) {
```

Could become:

```monkey
        } else if (item instanceof HomeAssistantMenuItem || item instanceof HomeAssistantIconMenuItem) {
            var haItem = item as HomeAssistantMenuItem or HomeAssistantIconMenuItem;
            if (Globals.scDebug) {
                System.println(haItem.getLabel() + " " + haItem.getId());
            }
            haItem.callService();
        } else if (item instanceof HomeAssistantViewMenuItem) {
```

Or does that not compile?

philipabbey: Euch! Just because you can factorise, doesn't mean you should.

We could collapse another pair of clauses too. Which is fine for now. But in the future when changes happen it will be easier to debug. The current expansion is more explicit.

JosephAbbey (approved):

PR #24, opened 2023-12-02
Initial version with "auto quit"

philipabbey:
Quit the application after a user settable period of time based on a timeout value from the settings.

@Someone0nEarth is this non-glance app of any interest to you?

### DIFF
--- a/resources/settings/settings.xml
+++ b/resources/settings/settings.xml
@@ -44,6 +44,16 @@
   </setting>
 
   <setting
+    propertyKey="@Properties.app_timeout"
+    title="Timeout in seconds. Exit the application after this period of inactivity to save the device battery."
+  >
+    <settingConfig
+      type="numeric"
+      min="0"
+    />
+  </setting>
+
+  <setting
     propertyKey="@Properties.types_representation"
     title="Representing types with icons (off) or with labels (on)"
   >
@@ -60,4 +70,12 @@
       type="boolean"
     />
   </setting>
+
+<!--
+  <defaults>
+    <entry>
+      <default id="app_timeout">20</default>
+    </entry>
+  </defaults>
+-->
 </settings>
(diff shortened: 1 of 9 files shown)


Someone0nEarth:
Hi!

This would prevent the battery drain issue. 👍

But on the UX level, having a widget/glance is much more comfy for me: Only to swipe down or up for starting. As an app, I would have to press a button on the watch (using my thumb and one finger) and then selecting the app for starting.

philipabbey:
I think we can have both.

- Testing has shown the widget still gives me what I want, a standard application, and what you want.
- We then add the auto quit so the app also exits.
   - This should either live along side the auto close of the widget, or be disabled (timeout=0).
 
Perhaps this combination gives us the best of all worlds?

Someone0nEarth:
That should work for me.

So, we add the timer you implemented and also change the type from "watch-app" to "widget" in the manifest.xml .

There is one "ugly" thing left in my dirt & quick widget/glance hack: When the user swipes back for quitting the widget, in the end, there will be a unnecessary "black screen" to swipe off. This is a circumstance of how this application views are currently implemented. I will try to change this later today or tomorrow and then we can see, if we will like it and go further to change the app to widget type.

philipabbey:
Suggest we close this one off as an interim solution that should be compatible (or at least tolerable) with the widget proposal from @Someone0nEarth.

resources/settings/settings.xml
```diff
@@ -60,4 +70,12 @@
       type="boolean"
     />
   </setting>
+
+<!--
```
JosephAbbey, line 74: It is probably not the best idea to leave stale code that does not work commented out.

JosephAbbey (changes requested):

JosephAbbey (approved):

PR #28, opened 2023-12-10
Auto close confirmation dialog after timeout

j-a-n:
Hello and thank you for your work on this project!
I have recently started using the app on my Garmin watch.

I find the confirmation function very important as I don't want to accidentally trigger actions like opening the front door.

To make the whole thing even more secure against accidental operation, I have added a timeout after which the confirmation dialog closes automatically. The timeout is currently hardcoded to 3 seconds, but can of course be configured if required.

 I hope you find this usefull as well.

### DIFF
--- a/source/HomeAssistantConfirmation.mc
+++ b/source/HomeAssistantConfirmation.mc
@@ -22,9 +22,9 @@ using Toybox.Lang;
 // Required for callback method definition
 typedef Method as Toybox.Lang.Method;
 using Toybox.WatchUi;
+using Toybox.Timer;
 
 class HomeAssistantConfirmation extends WatchUi.Confirmation {
-
     function initialize() {
         WatchUi.Confirmation.initialize(WatchUi.loadResource($.Rez.Strings.Confirm));
     }
@@ -33,17 +33,26 @@ class HomeAssistantConfirmation extends WatchUi.Confirmation {
 
 class HomeAssistantConfirmationDelegate extends WatchUi.ConfirmationDelegate {
     private var confirmMethod;
+    private var timeout;
 
     function initialize(callback as Method() as Void) {
         WatchUi.ConfirmationDelegate.initialize();
         confirmMethod = callback;
+        timeout = new Timer.Timer();
+        timeout.start(method(:onTimeout), 3000, true);
     }
 
     function onResponse(response) as Lang.Boolean {
         getApp().getQuitTimer().reset();
+        timeout.stop();
         if (response == WatchUi.CONFIRM_YES) {
             confirmMethod.invoke();
         }
         return true;
+    }
+
+    function onTimeout() as Void {
+        timeout.stop();
+        WatchUi.popView(WatchUi.SLIDE_RIGHT);
     }
 }
--- a/source/HomeAssistantConfirmation.mc
+++ b/source/HomeAssistantConfirmation.mc
@@ -23,6 +23,7 @@ using Toybox.Lang;
 typedef Method as Toybox.Lang.Method;
 using Toybox.WatchUi;
 using Toybox.Timer;
+using Toybox.Application.Properties;
 
 class HomeAssistantConfirmation extends WatchUi.Confirmation {
     function initialize() {
@@ -38,13 +39,18 @@ class HomeAssistantConfirmationDelegate extends WatchUi.ConfirmationDelegate {
     function initialize(callback as Method() as Void) {
         WatchUi.ConfirmationDelegate.initialize();
         confirmMethod = callback;
-        timeout = new Timer.Timer();
-        timeout.start(method(:onTimeout), 3000, true);
+        var timeoutSeconds = Properties.getValue("confirm_timeout") as Lang.Number; 
+        if (timeoutSeconds > 0) {
+            timeout = new Timer.Timer();
+            timeout.start(method(:onTimeout), timeoutSeconds * 1000, true);
+        }
     }
 
     function onResponse(response) as Lang.Boolean {
         getApp().getQuitTimer().reset();
-        timeout.stop();
+        if (timeout) {
+            timeout.stop();
+        }
         if (response == WatchUi.CONFIRM_YES) {
             confirmMethod.invoke();
         }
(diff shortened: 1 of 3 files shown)


source/HomeAssistantConfirmation.mc
```diff
@@ -33,17 +33,26 @@ class HomeAssistantConfirmation extends WatchUi.Confirmation {
 
 class HomeAssistantConfirmationDelegate extends WatchUi.ConfirmationDelegate {
     private var confirmMethod;
+    private var timeout;
 
     function initialize(callback as Method() as Void) {
         WatchUi.ConfirmationDelegate.initialize();
         confirmMethod = callback;
+        timeout = new Timer.Timer();
+        timeout.start(method(:onTimeout), 3000, true);
```
philipabbey, line 42: Do we want to move the `3000` ms into a Property that can be set by the user? We can default to 3 s (and multiply by 1000 after [`Properties.getValue()`](https://developer.garmin.com/connect-iq/api-docs/Toybox/Application/Properties.html#getValue-instance_function) as we've already done here [`(Properties.getValue("app_timeout") as Lang.Number) * 1000`](https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/source/QuitTimer.mc#L32)).

source/HomeAssistantConfirmation.mc
```diff
@@ -33,17 +33,26 @@ class HomeAssistantConfirmation extends WatchUi.Confirmation {
...
         getApp().getQuitTimer().reset();
+        timeout.stop();
         if (response == WatchUi.CONFIRM_YES) {
             confirmMethod.invoke();
         }
         return true;
     }
+
+    function onTimeout() as Void {
+        timeout.stop();
```
philipabbey, line 55: I believe this only needs to be called for repeating timers according to the docs at [`Timer.stop()`](https://developer.garmin.com/connect-iq/api-docs/Toybox/Timer/Timer.html#stop-instance_function).

But the call at line 47 needs to stay 😉.

philipabbey (changes requested):
> I find the confirmation function very important as I don't want to accidentally trigger actions like opening the front door.

I like you idea. I was worried people might use this for locks, but your addition mitigates that risk. I would like to make this the next update and can turn that around after these two comments are considered.

Thank you.

Philip

source/HomeAssistantConfirmation.mc
j-a-n: I thought about that too. I think that the 3 seconds is a good default and that it doesn't have to be configurable. I would wait for user feedback here.

source/HomeAssistantConfirmation.mc
j-a-n: I had assumed the same.
But without this statement, the app crashes after a few aborted confirmation dialogs.
Unfortunately, I'm not quite sure why.

source/HomeAssistantConfirmation.mc
j-a-n: If you disagree, I can implement that.

source/HomeAssistantConfirmation.mc
philipabbey: Ah, I wonder if that's because of some limit on the number of Timers allowed?

```
The number of available timers (default 3) and the minimum time value (default 50 ms) depends on the host system. An error will occur if too many timers are set.
```
(See top of the page at https://developer.garmin.com/connect-iq/api-docs/Toybox/Timer/Timer.html)

source/HomeAssistantConfirmation.mc
philipabbey: I tend to agree that you've tested 3 seconds is long enough, I'm just mindful that other users' mileage may vary. The change is really quite simple, as an example for the application timeout see:

- https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/resources/settings/properties.xml#L31
- https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/resources/settings/settings.xml#L46

So should be the work of a few minutes, if you don't mind?

Philip

j-a-n: No problem, I'll take care of it as soon as I have time.

philipabbey: We can also make the use of a `timeout` optional as previously done for the `QuitTimer`.

```monkey
if (api_timeout > 0) {
    start(method(:exitApp), api_timeout, false);
}
```
(See https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/source/QuitTimer.mc#L44.)

So that would be something like:

```monkey
        timeout = Properties.getValue("confirm_timeout");
        if (timeout > 0) {
            timer = new Timer.Timer();
            timer.start(method(:onTimeout), timeout, true);
        }
```

j-a-n: Or entry specific?

philipabbey: I did ponder this approach and initially thought it would be over the top. The schema would need another change:

```json
    {
      "entity": "script.food_on_table",
      "name": "Food is Ready!",
      "type": "tap",
      "tap_action": {
        "service": "script.turn_on",
        "confirm": true,
        "confirm_timeout": 3
      }
    },
```

That new field would need to permeate from `HomeAssistantView` through `HomeAssistantMenuItemFactory` all the way down. So there's a fair bit of code churn but doable. So the Properties based setting is a cheaper win, and does not preclude a per item configurable timeout in the future defaulting to the value in `Properties.getValue("confirm_timeout")`?

If you're up for it...? I think the single global setting is perhaps sufficient and mitigates others suffering our preferred timeout.

Anyway, I'm about to test your code as it stands on my watch.

j-a-n: I have now added the application property `confirm_timeout`.

j-a-n:
Thanks for the merge @philipabbey !

philipabbey:
Thank you for the contribution! I do have a user reported bug to fix before the next update goes out. A Vivoactive 4S crash has been reported.

PR #30, opened 2023-12-16
Change type to widget with glance view support

Someone0nEarth:


source/HomeAssistantGlanceView.mc
```diff
@@ -0,0 +1,40 @@
...
+// GarminHomeAssistant is a Garmin IQ application written in Monkey C and routinely
+// tested on a Venu 2 device. The source code is provided at:
+//            https://github.com/house-of-abbey/GarminHomeAssistant.
+//
+// P A Abbey & J D Abbey & SomeoneOnEarth, 23 November 2023
+//
+//
+// Description:
+//
+// Application root for GarminHomeAssistant
```
JosephAbbey, line 17: Should be "Glance view for GarminHomeAssistant" or something.

source/RootView.mc
```diff
@@ -0,0 +1,130 @@
...
+            :justification     => Graphics.TEXT_JUSTIFY_CENTER | Graphics.TEXT_JUSTIFY_VCENTER,
+            :locX              => 0,
+            :locY              => 83,
+            :width             => w,
+            :height            => h - 166
+        });
+    }
+
+    function onUpdate(dc as Graphics.Dc) as Void {
+        var bg = 0x3B444C;
```
JosephAbbey, line 87: Is it possible to replace this with `Graphics.COLOR_BLUE` for readability?
Is this a specific blue for a reason?

JosephAbbey (changes requested):
Just minor tweeks, this is looking good!

source/HomeAssistantGlanceView.mc
```diff
@@ -0,0 +1,40 @@
...
+
+    function initialize() {
+        GlanceView.initialize();
+    }
+
+    function onUpdate(dc) as Void {
+        GlanceView.onUpdate(dc);
+      
+        var font = Graphics.FONT_TINY;
+        var text = "GarminHomeAssistant";
```
philipabbey, line 35: How about we fetch the application name from the string resource file?
`WatchUi.loadResource($.Rez.Strings.AppName)`

philipabbey (changes requested):
Looking good 🕺.

source/RootView.mc
Someone0nEarth: > Is it possible to replace this with `Graphics.COLOR_BLUE` for readability? Is this a specific blue for a reason?

Copy pasta bg color from ErrorView. Was thinking to move this color to Globals.mc (but forgot it :)

philipabbey: I think that was my fault. I originally just copied the ErrorView.mc from another application. Would you mind sticking `Graphics.COLOR_BLUE` in to fix that legacy code please?

source/HomeAssistantGlanceView.mc
Someone0nEarth: https://github.com/house-of-abbey/GarminHomeAssistant/commit/e6d5053ebba4b2fa1d30ccadf7e549b96ea05ad4

source/RootView.mc
JosephAbbey: In fact, it may be better to use black here to be less jarring.

Someone0nEarth: https://github.com/house-of-abbey/GarminHomeAssistant/commit/ba5f90c682734b2570e8622c68a58a90919780ec

Change ErrorView to blue and RootView to black (error views should have their "own" color for better differentiating)

JosephAbbey:
Nice work!

Someone0nEarth:
Oh, there is still the widget uuid in the manifest.. should be change back to the original watch-app uuid in `main`, I guess

https://github.com/house-of-abbey/GarminHomeAssistant/blob/e3a6da4c9f1566ba5068166783f55bf45da7cbe1/manifest.xml

philipabbey:
Don't worry about the App ID in the manifest, we just keep changing it to match the test/live deployment route. Your private App ID will just be another in the comments for swapping in.

PR #35, opened 2023-12-21
34 compilation of widget application plus additional devices

philipabbey:
This should now build separate widget and watch-app .IQ files. There have also been some tweaks now that the two are separate so that the compromise made to widgets to keep the app users happy has been amended so that widgets act like other widgets and both the glance and widget root views contain some status.

@Someone0nEarth please will you review these changes? They will requite you to amend the app IDs to those you need and build via `export.cmd` which may need some paths customising to your needs.


Someone0nEarth:
> This should now build separate widget and watch-app .IQ files. There have also been some tweaks now that the two are separate so that the compromise made to widgets to keep the app users happy has been amended so that widgets act like other widgets and both the glance and widget root views contain some status.
> 
> 
> 
> @Someone0nEarth please will you review these changes? They will requite you to amend the app IDs to those you need and build via `export.cmd` which may need some paths customising to your needs.

Nice! I will try to do the review etc. tomorrow.

philipabbey:
So I note one problem. If your phone is dead and hence your watch can't communicate to the Home Assistant, there's an infinite loop it can't get out of, an error message about not being connected to the phone. Only way to get out is to reboot the watch. There's no overiding way to quite the application or widget.

philipabbey:
I note the memory limits on certain watches, e.g. we're testing the widget on an old Vivoactive3 (battery is knackered).

https://developer.garmin.com/connect-iq/reference-guides/devices-reference/#v%C3%ADvoactive%C2%AE3

image

Presently the app won't simulate as a widet for Vivoactive 3 without crashing unless compiled for release. Compiling for debug means its out of memory before it actually runs.

Just a note that we migth be reaching the limit of what we can include in this app for the future.

Someone0nEarth:
Hello!!

First review thingies:

- I am on Linux: I will try to provide a BASH version of `export.cmd` next week
- Widget Mode: The RootView not longer switch to the MenuView automatically (when connected successfully to HA). Instead its providing status information and the user have to interact to go further. I did the automated switch to the MenuView on purpose :) There is not really a benefit to have the status information in widget mode (Beside for the glance view: The status information can have a benefit). Have to interact to proceed lowers the UX. I would suggest, to switch back to the original RootView (plus automated switch) for Widget and keep the status view for the glance.
- For keep thingies cleaner and less complicated, I would suggest to remove `isWidget` and `switchToView` part, and switching back to the old `RootView` behavior: The "fetching" RootView is valid for watch-app type, too. And when stopping the app, the exit-RootView won`t be displayed for watch-app type/glance at all.

How is your way to test the widget version in the emulator using VSCode? I changed the type in the manifest.xml to widget and set the `isWidget` constant to "true". Is there a smoother way?

Someone0nEarth:
> So I note one problem. If your phone is dead and hence your watch can't communicate to the Home Assistant, there's an infinite loop it can't get out of, an error message about not being connected to the phone. Only way to get out is to reboot the watch. There's no overiding way to quite the application or widget.

Did`t had a deeper look into this yet. You mean, when the app is already connected to HA (and is showing the MenuView) and then connection to the phone dies?

Someone0nEarth:
> I note the memory limits on certain watches, e.g. we're testing the widget on an old Vivoactive3 (battery is knackered).
> 
> https://developer.garmin.com/connect-iq/reference-guides/devices-reference/#v%C3%ADvoactive%C2%AE3
> 
> image
> 
> Presently the app won't simulate as a widet for Vivoactive 3 without crashing unless compiled for release. Compiling for debug means its out of memory before it actually runs.
> 
> Just a note that we migth be reaching the limit of what we can include in this app for the future.

Okay. We can go for some code optimization and feature optimization, after we finished the upcoming bigger change. One possible feature optimization I could see, is to toggle on/off the toggle status updates and have a setting, how fast the toggle status should be updated.

philipabbey:
> The RootView not longer switch to the MenuView automatically (when connected successfully to HA). Instead its providing status information and the user have to interact to go further. I did the automated switch to the MenuView on purpose :)

Right, as this behaviour does not match other widgets on old devices, I assumed this was a compromise to keep the app working. My problem with the automatic starting of the app is the different user interface and hence look and feel.

> There is not really a benefit to have the status information in widget mode (Beside for the glance view: The status information can have a benefit).

I see no benefit from either the glance or the widget view having the status info, which was why i was happy to retain just the application on its own. But as stated above, widgets don't normally automatically start their apps, they wait until they are tapped.

The automatic start also prevents users from continuing to scroll throug their widgets if they are not quick enough. I guess that's why the other apps behave that way.

> I would suggest, to switch back to the original RootView (plus automated switch) for Widget

That's a UI busting move!

> and keep the status view for the glance.

I can live without the Glance, but if we must have it, it ought to be more than an App name!

> And when stopping the app, the exit-RootView won`t be displayed for watch-app type/glance at all.

All watches with widgets exit back to their widget view in order to allow users to scroll the carousel of widgets.

> How is your way to test the widget version in the emulator using VSCode? I changed the type in the manifest.xml to widget and set the isWidget constant to "true". Is there a smoother way?

Two changes as per comments in `manifest.xml` and `monkey.jungle`. There are a pair of lines to swap the comments on.

> You mean, when the app is already connected to HA (and is showing the MenuView) and then connection to the phone dies?

I've been chasing this one down today. I think I have fixed it. The problem was a race condition when the ErrorView was popped and pushed in close succession. The method of causing the problem was just one way. If you are out of Bluetooth range when going into the App/Widget is could lock up. Turning Bluetooth back on did not always fix the issue.

Okay, so I think we're down to a fundamental difference of opinion. You coded what you wanted for the Widget, but the behaviour is non-standard with all other widgets. That makes me nervous. Hence the work on the RootView to implement something more widget-like. **So we need to thrash this one out!**

philipabbey:
> Okay. We can go for some code optimization and feature optimization, after we finished the upcoming bigger change. One possible feature optimization I could see, is to toggle on/off the toggle status updates and have a setting, how fast the toggle status should be updated.

The speed of the updates won't change the overall application memory footprint much if at all. The present implementation does not have a controllable speed as they are chanined from the last. We can slow this down of course, but that makes the application less responsive.

In order to grip the memory usage, we just need to not add any more code! I've already found a case where removing a member variable (bitmap icon) in favour of a local variable in a method seemed to solve an out of memory issue.

Someone0nEarth:
> > The RootView not longer switch to the MenuView automatically (when connected successfully to HA). Instead its providing status information and the user have to interact to go further. I did the automated switch to the MenuView on purpose :)
> 
> Right, as this behaviour does not match other widgets on old devices, I assumed this was a compromise to keep the app working. My problem with the automatic starting of the app is the different user interface and hence look and feel.

Right, the "normal" way is to keep the carousel running.. . Have the watch now for 3 years.. and true, I don`t like, how the whole widget-carousel behavior is designed. For me, swiping to left to exit a widget is fine, I don`t need to carousel after I am done with a widget. But you are right, its better to keep the original design behavior. I could see a setting, where the user can activate the automated switch to the MenuView.

> I can live without the Glance, but if we must have it, it ought to be more than an App name!

There glances with only the App name.. just for starting the App.. but showing the status information is ok. May can toggle via a setting?
 
> > How is your way to test the widget version in the emulator using VSCode? I changed the type in the manifest.xml to widget and set the isWidget constant to "true". Is there a smoother way?
> 
> Two changes as per comments in `manifest.xml` and `monkey.jungle`. There are a pair of lines to swap the comments on.

Ah, I oversaw this ✌️

> 
> > You mean, when the app is already connected to HA (and is showing the MenuView) and then connection to the phone dies?
> 
> I've been chasing this one down today. I think I have fixed it. The problem was a race condition when the ErrorView was popped and pushed in close succession. The method of causing the problem was just one way. If you are out of Bluetooth range when going into the App/Widget is could lock up. Turning Bluetooth back on did not always fix the issue.

👍️

 
> Okay, so I think we're down to a fundamental difference of opinion. You coded what you wanted for the Widget, but the behaviour is non-standard with all other widgets. That makes me nervous. Hence the work on the RootView to implement something more widget-like. **So we need to thrash this one out!**

Hehe, yeah and we will do :)

My proposals:

- Keeping status information for glance and widget
- Setting for the glance to show only the App name (could be something for the future)
- Setting for widget to automatically switch to the MenuView

What do you think? I could implement these changes.

philipabbey:
> There glances with only the App name.. just for starting the App.

I've noticed, it does miss the point! So I tend not to have those wasting space in my carousel...

> My proposals:
> 
> - Keeping status information for glance and widget
> - Setting for the glance to show only the App name (could be something for the future)

For me, either:
- Glance with more than App name, or
- Start as an App which already has the app name.

We're in danger of options for options sake! The extra status info can act as an alert to not having configured the settings correctly. My latest (as yet unpushed) changes do better at that.

> - Setting for widget to automatically switch to the MenuView

I can live with that, but I would suggest the default shoudl match the expected behaviour.

> What do you think? I could implement these changes.

I'll pile that last one in with my current changes still being tested.

Someone0nEarth:
> For me, either:
> 
>     * Glance with more than App name, or
> 
>     * Start as an App which already has the app name.

I don`t get the second one :)

> We're in danger of options for options sake! The extra status info can act as an alert to not having configured the settings correctly. My latest (as yet unpushed) changes do better at that.

Yep, was only thinking about battery consumption, for keeping the glance status up to date. So, maybe, some users would be more happy with App name only. But as I wrote, could be something for the future, if this would be really thingy.

> 
> > * Setting for widget to automatically switch to the MenuView
> 
> I can live with that, but I would suggest the default should match the expected behaviour.

Yep, default behavior should be the "normal" one.

> I'll pile that last one in with my current changes still being tested.

Okay! When ready for testing and review, just ping me :)

philipabbey:
> Yep, was only thinking about battery consumption, for keeping the glance status up to date.

The glance status is fine on a timer, 1s presently. One does not linger long on the glances anyway.

philipabbey:
> > For me, either:
> > ```
> > * Glance with more than App name, or
> > 
> > * Start as an App which already has the app name.
> > ```
> 
> I don`t get the second one :)

Screenshot from Venu2:

image

Photo of Vivoactive 3:

image

Both show the app as one buttom press away from the watch face with its name next to the icon, so very accessible! In both cases I just marked the app as a favourite.

philipabbey:
@Someone0nEarth please can you pull and check this latested commit.

JosephAbbey (approved):

Someone0nEarth:
Happy new year!

I was not be able to do stuff in the last week. I ll try to write the Linux bash-script this week.

PR #36, opened 2023-12-28
Added an option to enable a secondary glance design

avalentic:
I have made some changes to the glance because the old design didn't fit in with all other glances from other apps, and I didn't like that (although the whole app is excellent!).
That is why I created this feature, where you can change the setting `glance_secondary_design,` which (if enabled) will remove the blue background from the glance and make left alignment for all text at a glance (with a little smaller left margin, too).
Here is a screenshot of the glance on Fenix 7S Pro (emulated) with the setting disabled.
Screenshot 2023-12-28 at 10 44 19
The screenshot of a glance with the setting enabled
Screenshot 2023-12-28 at 10 39 56

This is my first contribution to some Connect IQ apps, and I also don't have much professional C experience, so I'm open to any suggestions on improving this feature! Happy holidays!

### DIFF
--- a/resources/settings/properties.xml
+++ b/resources/settings/properties.xml
@@ -54,4 +54,9 @@
   -->
   <property id="widget_start_no_tap" type="boolean"></property>
 
+  <!--
+    Enable or disable glance background and right alignment
+  -->
+  <property id="glance_secondary_design" type="boolean"></property>
+
 </properties>
--- a/resources/settings/settings.xml
+++ b/resources/settings/settings.xml
@@ -90,4 +90,13 @@
     />
   </setting>
 
+    <setting
+    propertyKey="@Properties.glance_secondary_design"
+    title="@Strings.SettingsGlanceSecondaryDesign"
+  >
+    <settingConfig
+      type="boolean"
+    />
+  </setting>
+
 </settings>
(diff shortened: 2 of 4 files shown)


philipabbey:
Hi @avalentic,

Thanks for you interest and feedback.

I won't be creating a secondary glance, but I'm happy to adopt your design. Too many options...
I was trying to be true to Home assistant colours, but I'm not bothered enough about my design to keep it. The margin however I recall has proven significant with different devices so I will double check that can be removed.

avalentic:
@philipabbey should I create a new pull request without the option or you will just copy the code?

Also what do you think about updating a Home Assistant logo?

philipabbey:
Venu2 glance

Done. Changes were trivial. I did alignt the second column of text though.

> Also what do you think about updating a Home Assistant logo?

I believe we're using their latest logo already, see https://www.home-assistant.io/blog/2023/09/17/a-refreshed-logo-for-home-assistant/.

avalentic:
Looks amazing, thank you @philipabbey

And I think we are using the first logo from 2013, the latest one can be found here
https://github.com/home-assistant/assets/tree/master/logo

philipabbey:
> And I think we are using the first logo from 2013, the latest one can be found here https://github.com/home-assistant/assets/tree/master/logo

Argh! The date in the URL suggests I'm using the logo from 17 Sep 2023, except they displayed the old logo prominently at the top. I assumed they were just playing catchup.

PR #53, opened 2024-01-10
Use webhooks for reporting battery

JosephAbbey:


source/BackgroundServiceDelegate.mc
```diff
@@ -51,17 +51,26 @@ class BackgroundServiceDelegate extends System.ServiceDelegate {
         } else {
             // Don't use Settings.* here as the object lasts < 30 secs and is recreated each time the background service is run
             Communications.makeWebRequest(
-                (Properties.getValue("api_url") as Lang.String) + "/events/garmin.battery_level",
+                (Properties.getValue("api_url") as Lang.String) + "/webhook/" + (Properties.getValue("webhook_id") as Lang.String),
```
philipabbey, line 54: Could just use a static const in Globals.mc

source/Settings.mc
```diff
@@ -25,13 +25,118 @@ using Toybox.System;
...
+                    "state"           => System.getSystemStats().charging,
+                    "type"            => "binary_sensor",
+                    "unique_id"       => "battery_is_charging",
+                    "entity_category" => "diagnostic",
+                    "disabled"        => false
+                });
+            }
+        } else {
+            if (Globals.scDebug) {
+                System.println("Settings onReturnRequestWebhookId(): Error: " + responseCode);
```
philipabbey, line 60: WebhookManager rather than 'Settings'. And all that follow too.

source/Settings.mc
```diff
@@ -25,13 +25,118 @@ using Toybox.System;
...
+    function requestWebhookId() {
+        if (Globals.scDebug) {
+            System.println("Settings requestWebhookId(): Requesting webhook id");
+        }
+        Communications.makeWebRequest(
+            Settings.getApiUrl() + "/mobile_app/registrations",
+            {
+                "device_id"           => System.getDeviceSettings().uniqueIdentifier,
+                "app_id"              => "garmin_home_assistant",
+                "app_name"            => "GarminHomeAssistant",
```
philipabbey, line 74: RezStrings.getAppName(). Might need fishing out of strings.xml.

source/Settings.mc
```diff
@@ -25,13 +25,118 @@ using Toybox.System;
 // Battery Level Reporting
 using Toybox.Background;
 using Toybox.Time;
+using Toybox.Communications;
+
+class WebhookManager {
```
philipabbey, line 30: Separate file to Settings?

source/BackgroundServiceDelegate.mc
philipabbey: webhook_id from Properties could be static const in Globals perhaps?

JosephAbbey: But it changes when `WebhookManager::requestWebhookId()` is run. I have also made it a actual seetting as part of the trouble shooting steps could be to create it yourself and paste it in.

philipabbey: Doh!

philipabbey: > I have also made it a actual seetting as part of the trouble shooting steps could be to create it yourself and paste it in.

The need for debug would preclude using `Settings` as opposed to `Properties` then.

JosephAbbey:
I don't know the best way to handle errors when registering the sensors (stage 2).

Possible ways:

- Keep track of whether it was successful and try again next time (reset when the webhook id is changed).
- Prompt the user to do it manually with a script
- Automatically do it when that [error](https://developers.home-assistant.io/docs/api/native-app-integration/sensors/#updating-a-sensor) is received in the background service.

> If an update was unsuccessful, an error is returned.
> 
> ```json
> {
>   ...
>   "battery_charging": {
>     "success": false,
>     "error": {
>       "code": "not_registered",
>       "message": "Entity is not registered",
>     }
>   },
>   ...
> }
> ```

@philipabbey do you have a preference either way?

philipabbey:
We decided that if either of the secondary setup requests to make devices fails, then to back back out as an overall failure, resetting the webhook property. If the failure was temporary, don't unset the battery level update setting. If more serious failure, unset the batter update setting so that we are not stuck in an error loop.

source/Settings.mc
```diff
@@ -131,4 +155,11 @@ class Settings {
         return mIsWidgetStartNoTap;
     }
 
+    static function disableBatteryLevel() {
```
philipabbey, line 158: Darn, I get the context of the question earlier now. Previously I used `unset` rather than `disable` for `static function unsetClearCache()`. So we have inconsistency, but I'm minded to change the original one to match this.

PR #72, opened 2024-01-14
Template Menu Item

JosephAbbey:
Runs on vivoactive3 without memory issues


source/Globals.mc
```diff
@@ -24,7 +24,7 @@ using Toybox.Lang;
 class Globals {
     // Enable printing of messages to the debug console (don't make this a Property
     // as the messages can't be read from a watch!)
-    static const scDebug        = false;
+    static const scDebug        = true;
```
philipabbey, line 27: Needs resetting to `false`.

source/RezStrings.mc
```diff
@@ -100,6 +101,7 @@ class RezStrings {
         strUnhandledHttpErr  = WatchUi.loadResource($.Rez.Strings.UnhandledHttpErr);
         strTrailingSlashErr  = WatchUi.loadResource($.Rez.Strings.TrailingSlashErr);
         strWebhookFailed     = WatchUi.loadResource($.Rez.Strings.WebhookFailed);
+        strTemplateError     = WatchUi.loadResource($.Rez.Strings.TemplateError);
```
philipabbey, line 104: Need to re-run automatic translation script.

philipabbey (changes requested):
Looking good. About to compile it up for test deployment.

We need a page on examples of how the JSON configuration can now be exploited. Give people some ideas.

philipabbey:
Further testing is revealing memory limits. The widget fails when the application works. The simulator shows us the memory limits as follows.
- Vivoactive3 App 124kB
- Vivoactive3 Widget 60kB

The widget is now using 56 of 59.9 kB and crashing.
(Game not over yet)

philipabbey:
Using a script to ensure compilaion of Vivoactive 3 without debug (--release) and deploy to the simulator we have 52.4/59.9 kB used.

Changing the JSON configuration to a simpler one (sill using no debug) we have 49.0/59.9 kB used.

So we have confirmation that the JSON configuration can cause this device to run out of memory. Not an unexpected result, but we are clearly "on the edge".

philipabbey:
Latest commits provide a saving by removing the menu style option and associated class:

Vivoactiove 3 Widget with template menu items (previously failing on watch) now 50.9/59.9 kB

philipabbey:
Latest commit now 50.4/59.9 kB

philipabbey (approved):

PR #76, opened 2024-01-14
74 cache has results

philipabbey:
The `has` calls have been cached with little effect on memory usage.
Memory usage has been added to the RootView so that widgets (where memory is more limited than applications) can see how close they are to running out.
Documentation updates previously missed has been addressed.


philipabbey:
I have rejected README styling changes. They were unhelpful suggestions for tables, particularly where the table rows were multi-line. Please disregard or even better turn markdown checks off!

JosephAbbey (approved):

PR #82, opened 2024-01-18
52 allow parameters to be send with tap menu items

philipabbey:
Calling for review. Just testing the effect of removing `ErrorView.unShow()` from callbacks on actual devices.


source/HomeAssistantView.mc
```diff
@@ -51,11 +51,13 @@ class HomeAssistantView extends WatchUi.Menu2 {
...
-            var service    = items[i].get("service")    as Lang.String     or Null;
-            var confirm    = false                      as Lang.Boolean;
+            var service    = items[i].get("service")    as Lang.String     or Null; // Deprecated schema
+            var confirm    = false                      as Lang.Boolean    or Null;
+            var data       = null                       as Lang.Dictionary or Null;
             if (tap_action != null) {
                 service = tap_action.get("service");
-                confirm = tap_action.get("confirm");
+                confirm = tap_action.get("confirm"); // Optional
+                data    = tap_action.get("data");    // Optional
```
JosephAbbey, line 60: Should these have type annotations like the ones above?

source/HomeAssistantMenuItem.mc
```diff
@@ -48,9 +50,14 @@ class HomeAssistantMenuItem extends WatchUi.IconMenuItem {
...
         mHomeAssistantService = haService;
-        mIdentifier           = identifier;
         mService              = service;
         mConfirm              = confirm;
+        mData                 = data;
+        if (mData == null) {
+            mData = {"entity_id" => identifier};
+        } else {
+            mData.put("entity_id", identifier);
+        }
```
JosephAbbey, line 60: Do we have plans to conditionally include this to allow for services like notify that complain?

Should we be using `entity_id` as the identifier? If we use something else, `entity_id` can be optional which would solve this (especially as it isn't actually used for `TemplateMenuItem`).

JosephAbbey (changes requested):

source/HomeAssistantMenuItem.mc
philipabbey: So our previous discussion considered how we many exceptions to including `entity_id`, and we left it as, keep it simple, no exceptions. But we do have the opportunity, and we could make the `mIdentifier` a simple one-up number so that multiple menu items can use the same `entity_id`.

How many exceptions do we need? Since they have to be hard coded.

source/HomeAssistantView.mc
philipabbey:
```
            var service    = items[i].get("service")    as Lang.String     or Null; // Deprecated schema
            var confirm    = false                      as Lang.Boolean    or Null;
            var data       = null                       as Lang.Dictionary or Null;
```

> the ones above?

...are type definitions. Because they provide the type of the variable on the left hand side of these assignments, the "as Lang.*" in the assignment is already covered and becomes unnecessary surely?

source/HomeAssistantMenuItem.mc
philipabbey: @JosephAbbey > Should we be using entity_id as the identifier?
@philipabbey > we could make the mIdentifier a simple one-up number

The `HomeAssistantToggleMenuItem` continues to use `mIdentifier` in the `makeWebRequest()` call. Those object instances presently do not include any additional `data` field for parameters to populate with the current value of `mIdentifier`. So we would end up making an inconsistent change if performed this change for `toggle`, `template` and `tap`, or we need to add more code and a member variable to `HomeAssistantToggleMenuItem`, which bloats the code.

JosephAbbey (approved):

PR #88, opened 2024-01-22
87 look for memory usage efficiencies

philipabbey:
1kB savings!


JosephAbbey (approved):

philipabbey:
Stating the obvious perhaps, but we must also remember that caching a menu also takes up memory, so increasing the risk that the Widget version crashes out of memory.

PR #110, opened 2024-02-05
Fixed off-by-one error in Templates.md

mymyke:
Round function added the brightness display part of content to correct off-by-one errors. For example, setting brightnes to 30% would previously show 29% and now shows 30%


JosephAbbey (approved):
Hi @mymyke,

This is a nice addition to the documentation, we really appreciate it.

Thank you very much!

*.. Joseph*

PR #123, opened 2024-03-21
103 possiblity to add polling interval for states to save battery

philipabbey:
I managed the preferred solution, update all, pause, update all, pause etc. This will be version 2.10.


source/HomeAssistantApp.mc
```diff
@@ -393,15 +394,24 @@ class HomeAssistantApp extends Application.AppBase {
...
+            mUpdateTimer.start(method(:updateNextMenuItemInt), delay, false);
+        } else {
+            updateNextMenuItemInt();
+        }
+    }
+
     // We need to spread out the API calls so as not to overload the results queue and cause Communications.BLE_QUEUE_FULL
     // (-101) error. This function is called by a timer every Globals.menuItemUpdateInterval ms.
-    function updateNextMenuItem() as Void {
+    function updateNextMenuItemInt() as Void {
```
JosephAbbey, line 408: This naming threw me off at first, I assume it means internal? May be better to have the full word or just prefix it with an underscore (`_updateNextMenuItem`). Originally I thought it was int as in integer.

source/Settings.mc
```diff
@@ -36,6 +36,7 @@ class Settings {
     private static var mClearCache            as Lang.Boolean = false;
     private static var mVibrate               as Lang.Boolean = false;
     private static var mAppTimeout            as Lang.Number  = 0;  // seconds
+    private static var mPollDelay             as Lang.Number  = 0;  // milliseconds
```
JosephAbbey, line 39: Everywhere else says seconds, not milliseconds

JosephAbbey (changes requested):

JosephAbbey (approved):

JosephAbbey (approved):

PR #158, opened 2024-07-25
Update HomeAssistantTemplateMenuItem.mc

philipabbey:
Template errors now displayed per item.


philipabbey:
Need to internationalise the string "Template Error".

philipabbey:
Causes infinite loop as `mNextItemToUpdate` does not get incremented until after getState() returns. This all runs in the same thread.

```
        } else if (Settings.getWebhookId().equals("")) {
            getApp().updateNextMenuItem();
```

JosephAbbey (approved):

PR #170, opened 2024-08-25
165 templating content for all types

philipabbey:
Submitting as an initial draft in order to guage feedback on the style. In particular the solution to adding the code to the `toggle` menu item, which has incurred duplication of code as I can't inherit from two classes.


philipabbey:
Reminder to self: Need to update the examples to show the optional use of a "content" field.

config.schema.json
```diff
@@ -213,6 +218,10 @@
       },
       "required": ["service"]
     },
+    "content": {
```
JosephAbbey, line 221: Reuse of the `content` keyword means that the web editor will keep working. 👍

source/HomeAssistantToggleMenuItem.mc
```diff
@@ -272,4 +281,127 @@ class HomeAssistantToggleMenuItem extends WatchUi.ToggleMenuItem {
...
+                break;
+
+            default:
+                // System.println("HomeAssistantTemplateMenuItem onReturnGetState(): Unhandled HTTP response code = " + responseCode);
+                ErrorView.show(WatchUi.loadResource($.Rez.Strings.UnhandledHttpErr) as Lang.String + responseCode);
+        }
+        getApp().setApiStatus(status);
+    }
+
+    // Massive code duplication from TemplateMenuItem, but cannot inherit from two classes.
```
JosephAbbey, line 364: Could potentially be extracted into a function in a static class that takes callback, etc. as parameters?

philipabbey: I tried factorising it the template code to its own class. It got messy, as the call back requires access to the menu item's methods. I would have needed to pass several menu item methods as parameters, e.g subMenuTitle(), that it became a joke, so I ditched the code as duplication was leds messy.

config.schema.json
philipabbey: Okay, that was a thumbs up rather than a code change request.

source/HomeAssistantToggleMenuItem.mc
philipabbey: We agreed this code was awkward but necessary. It might all be negated by the next feature update anyway, so merging in.

JosephAbbey (approved):

PR #183, opened 2024-09-05
182 add support for the latest devices released from garmin

philipabbey:
7 new devices. 2 users waiting for this update for Fenix 8.


source/ErrorView.mc
```diff
@@ -91,7 +91,7 @@ class ErrorView extends ScalableView {
         return mDelegate;
     }
 
-    static function create(text as Lang.String) as Lang.Array<ErrorView or ErrorDelegate> {
+    static function create(text as Lang.String) as [ WatchUi.Views ] or [ WatchUi.Views, WatchUi.InputDelegates ] {
```
JosephAbbey, line 94: We should be consistent with the syntax for list types, either:

```monkeyc
[...]
```

or

```monkeyc
Lang.Array<...>
```

JosephAbbey (changes requested):

source/ErrorView.mc
philipabbey: We could be but...

1. They changed the function prototype on us. And inconsistently too.
2. Experience so far indicates `[...]` is not considered quiet the same as `Lang.Array<...>` by the compiler. So we are forced to match what they have used in the updated API.

I propose we just much their Connect IQ API syntax. If they use `Lang.Array` so do we, if they use `[...]` so do we. I would like to avoid churning all our code.

JosephAbbey (approved):

PR #186, opened 2024-09-07
2 fixes as required by issue 185

philipabbey:
Solution to a race condition with starting updates and amendment to error message presentation on Webhook ID creation failure.


philipabbey:
I have already reviewed your race condition changes. So that should just leave the trivial `WebhookManager.mc` file changes I made.

philipabbey:
This will not form a new release on its own, we'll just store it up for something bigger to come.

JosephAbbey (approved):

PR #192, opened 2024-10-16
Feature: PIN confirmation dialog

moesterheld:
## About this PR:
Being one of those incorrigible fools who want to open and unlock stuff using my watch, I thought of a way to improve security when calling a HomeAssistant service from the Garmin App.

This change adds a PIN confirmation dialog that is displayed whenever a `pin` property has been set in the `tap_action`. Because of the size of the screen, PINs consist of a sequence of abritrary length using the digits 1-4.

2024-10-16 11_16_09-CIQ Simulator - Venu® Sq 2 Music (5 0 0)

Due to public visibility of the dashboard config, a pin mask can be configured in the app's settings to transcode the PIN on the watch for additional security. Using a cipher for PIN encryption has been discarded due to the complexity involved in generating the encrypted PINs for the end user.

The PIN dialog is configured to allow up to 4 erraneous PIN inputs in 2 minutes, a 5th failure will lock the PIN dialog (for all actions) for 10 minutes. Entering a correct PIN will reset the failure count.

## Changes included in this PR:
- PIN confirmation screen
	- called whenever a `pin` has been configured in a `tap_action`. This overrides the `confirm` property. Users without a touchscreen will be routed to the regular confirmation dialog.
	- optional `pin_mask` configuration setting, which will be used to transcode the digits of the PIN (see documentation change)
	- german translation of the `pin_mask` label
	- failed attemps and PIN lock are stored in application storage
- Addition of `pin` field to the JSON Schema
- Additions to the docs

Tested on a few devices in the simulator, tested and running as a beta on a Venu 2 Sq Music.

## Disclaimer:
This is my first attempt at a Monkey C project.
I probably learned more from looking at the existing app than from the sdk samples. Awesome work! 

I added this mainly because I wanted a feature like that for myself, thought it is a good thing to add and because I was interested in developing for Garmin devices.
I would be glad if you like the idea and are open to review it and include it in the app.

I have tried to use coding style and conventions from the existing project as far as I could see them.

### DIFF
--- a/source/HomeAssistantApp.mc
+++ b/source/HomeAssistantApp.mc
@@ -111,7 +111,7 @@ class HomeAssistantApp extends Application.AppBase {
             return ErrorView.create(WatchUi.loadResource($.Rez.Strings.NoConfigUrl) as Lang.String + ".");
         } else if (Settings.getPin() == null) {
             // System.println("HomeAssistantApp getInitialView(): Invalid PIN in application settings.");
-            return ErrorView.create(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String);
+            return ErrorView.create(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String + ".");
         } else if (! System.getDeviceSettings().phoneConnected) {
             // System.println("HomeAssistantApp getInitialView(): No Phone connection, skipping API call.");
             return ErrorView.create(WatchUi.loadResource($.Rez.Strings.NoPhone) as Lang.String + ".");


philipabbey:
Apologies, only just seen this. I'll take a closer look this week.

philipabbey:
I have got this working in a simulation of the Venu2 device with a menu item that has a PIN. That's a good start! It's actually an impressive development over all. I confess I was initially sceptical, so I'm delighted the non-touch screen devices have been considered. As this is a major update I was going to deal with a round of general comments first before going into smaller code review details. I'm feeling positive about including this development. Although it does encourage people to use the GHA app for security sensitive devices, which I'm slightly concerned about. Very much "at your own risk", so again need to consider how this is presented having previously told people not to do it...

I think the PIN + mask feels complicated. I realise the PIN is public in the menu in clear, and hence you used a private transcode integer mask. But I wonder if it would be better to have a single PIN in the local watch's settings which is not public. That would be a single PIN for all menu items, and the JSON file then instead has an optional Boolean switch for "pin". Just feels simpler over all, and I'm not sure what separate PINs per menu item adds in value.

```json
 "tap_action": {
    "service": "cover.toggle",
    "pin": "true"
  }

```

I fear support trouble ahead for this present PIN + mask scheme.

Now I can quickly imagine how many users will set the transcode mask to "0000" as its annoying to reverse-transcode the PINs to store. Leaving their PINs in the clear and protected only by obscurity. The private PIN in the settings instead of the transcode value feels like the same level of security, since the only secure value presently is the transcode value in the private settings, which is no more secure than a single private PIN.

The PIN is too easy to reverse engineer because you fail the user at the first wrong digit. As a minimum, take all 4 digits before checking pass/fail.

Algorithm to break the PIN is:

1. Iterate 1-4 for first digit until it does not fail.
2. Enter first digit (now known) and iterate 1-4 for 2nd digit until it does not fail.
3. Enter first 2 digits (now known) and iterate 1-4 for 3rd digit until it does not fail.
4. Enter first 3 digits (now known) and iterate 1-4 for 4th digit until it does not fail.

Now we have the PIN in no more than 16 guesses. I realise you have a timeout feature, 5 tries in 2 mins. So it will only take 30 mins (worst case) to break in!
Instead, if the whole PIN had been read before checking it would take 4^4 = 256 guesses to exhaust. So we have a limited amount of PIN entropy (understood why) weakened by a vulnerability in the PIN entry method.

## Questions:

These are me thinking out loud, looking for points of view.

1. Would the PIN entry method you are using scale (on screen) to a 3x3 grid for an entropy of 9^4 = 6,561 guesses to exhaust? Feels more acceptably secure. It might also allow for a relaxation of the timeouts?

2. If use of a PIN is specified for a non-touch screen device, do we silently fall back to the confirmation option (as you have done), or should we actually print an error message to be explicit and force the user to remove the PIN option from their menu. In this way we manage expectations and don't get users asking why the PIN is not working?

## More trivial details for later:

Cosmetic: I think the digits placement could be adjusted to be more centred. No idea if this is a simulation only issue. This is a screen shot from a Venu 3 device.

PIN

Documentation will need an explanation of the various timeouts used for failed PINs. Those values are hard coded, which does not bother me immediately, but they would be better placed in the `Globals.mc` class.

PIN lockout - Needs internationalisation. Just amend the default resource strings in English, there's a Python script to translate to the other languages. Feel free to provide customised translations in `corrections.xml` for your preferred language.

Commented out debug - not a problem. The printed messages would benefit from class name plus function name for when they are used in a mix of other messages.

**Thank you for your contribution and interest in this project!**

moesterheld:
Thank you for the exhaustive review!

Considering your points:
I agree that adding the PIN globally is a good idea and would make it less error-prone. I discarded the same idea early-on, in order to allow for different pins for different devices. But you are right, that might have been over-the-top. I am all for it. We could even think about using the existing JSON and use the PIN for every `confirm: true`.

Good catch regarding the PIN validation. That needs to be changed.

Regarding the grid. I initially tried a 3x3 grid on my watch, but that was not very comfortable to work with, so I changed it to 2x2.  In the JSON schema I have not set a minimum or maximum length for the PIN, which leaves it up to the user to choose between safety and comfort. My pins are all 6 digits long so they have a 4^6 entropy. But let me see if I can find my initial attempts and push them so you can try it out.
When moving the PIN to a global setting, we should enforce a (fixed? minimum?) length. It looks like there is no easy way for validating the settings except for the very basic properties that are provided and do not fit, because I wanted to add that to the pin mask, too. Any idea? Maybe an error message on startup?

Layout: I will try to tweak this a bit more. Since I used textarea to automatically set the font size according to the bounding box, I did not want to look into how to dynamically move the text in the center of the square. Also, for round displays, that is probably not ideal, the text should be closer to the cross (hence the margin). I will dig into this. I noticed that I need to change background/foreground colors to reflect the watch's settings as well, but didn't want to add to the PR before you took a look at it.

PIN lockout i18n - you're right, I missed that, will add it

Debug messages - I will adapt the strings to include class/method info

I will probably find some time to address the issues next week. If you have any comments beforehand, feel free to let me know.

moesterheld:
also, 3x3 grids don't work too well on round watches

philipabbey:
> We could even think about using the existing JSON and use the PIN for every confirm: true.

Yes, agreed. That's what I intended.

Regarding round displays, the screen capture I took was from the simulator menu File -> Save Screen Capture. It does not include the round watch face. Here's another try.

image

Its all a bit fiddly I know.

2x2 vs 3x3: I was hoping this might be possible since these calculator apps use higher density grids, e.g. 4x4 or 5x3.
- https://apps.garmin.com/apps/5aec26ce-9e57-4bee-acb3-11c841b75e3f
- https://apps.garmin.com/apps/96ee6eae-798a-45b6-adb9-f569869d3d23
- https://apps.garmin.com/apps/72f4a243-e138-4e3b-91b2-404b394cd8b7

Chatting to my co-author @JosephAbbey:

> When moving the PIN to a global setting, we should enforce a (fixed? minimum?) length.

1. We agree that 4^6 = 4096 is close enough to 9^4 entropy (2/3rd the way there...). We also think the number of digits _must_ be fixed otherwise you cannot know how many digits to wait for before checking the PIN is correct.

2. You are right, constraining the user specified [PIN in the options](https://developer.garmin.com/connect-iq/core-topics/properties-and-app-settings/) is a pain. Only option is a `string` and only `maxLength` can be used to constrain it.

>Maybe an error message on startup?

Yes, better than at the time as needing to check the PIN. So validated in the `Settings.mc` class, with a blue screen of error. Agreed that's not the best, but we came to the same conclusion.

3. New vulnerability. If the app times you out from trying PINs, just exit and enter the app again to evade the timeout. Continue spoofing the PIN again. So we think that can be solved by using the [Storage](https://developer.garmin.com/connect-iq/api-docs/Toybox/Application/Storage.html) class to save state between runs like we already use for the menu cache. Either store the absolute time since epoch, or the relative left to wait out. When opening the app, pick up the old values and continue counting down. We then thought about how the timeout might be changed in the app settings to evade this, and decided only the watch's owner would likely have access to the settings, so this was not a security worry.

4. User feedback for how many digits entered (especially if 6 rather than 4). Options are some or all of:
   - Hepatic feedback (vibrate on digit entered)
   - progress bar (square for each digit gets filled in for each press), or
   - digit counter,
   - flash the background of the selected digit (hmm might be involved)

I would hope that would be an easyish addition to your new PIN view. (?!)

>The number of available timers (default 3) and the minimum time value (default 50 ms) depends on the host system. An error will occur if too many timers are set.

Worth noting that some watches are limited in the number of [Timers](https://developer.garmin.com/connect-iq/api-docs/Toybox/Timer/Timer.html) available. I'm not sure how you find out the maximum number. So some care needs to be taken with using yet another Timer.

JosephAbbey:
> 3. New vulnerability. If the app times you out from trying PINs, just exit and enter the app again to evade the timeout. Continue spoofing the PIN again. So we think that can be solved by using the [Storage](https://developer.garmin.com/connect-iq/api-docs/Toybox/Application/Storage.html) class to save state between runs like we already use for the menu cache. Either store the absolute time since epoch, or the relative left to wait out. When opening the app, pick up the old values and continue counting down. We then thought about how the timeout might be changed in the app settings to evade this, and decided only the watch's owner would likely have access to the settings, so this was not a security worry.

In this case it is better to use the timestamp, rather than storing the remaining time as this allows the timeout to continue when the app is closed. If the remaining time is stored, when you enter the wrong pin too many times you must keep the app open for the full timeout; so if you enter the wrong pin and get timed out, close the watch app to just use your phone or however else, then when you open the app again you will still have the full timeout remaining. Which is just infuriating for the end user.

moesterheld:
Thanks for the additional great feedback

https://github.com/house-of-abbey/GarminHomeAssistant/blob/cd5ed317c51f1aabf5f7ab4da72f047316019f95/source/HomeAssistantPinConfirmation.mc#L239

Here, I already store the failure timestamps in the Storage and use these. Unless I am missing something, I think I have that covered.

Additional to what we discussed before, as proposed in the last comment, I will: 
- optimize the grid (maybe even bigger), 
- user feedback (haptic feedback is already implemented, but will try to add an additional indicator) 
- add the "blue screen" setting validation
and get back to you. 

As for the timers, I will have to read up on that.

philipabbey:
> Here, I already store the failure timestamps in the Storage and use these. Unless I am missing something, I think I have that covered.

You do, I overlooked that.

> haptic feedback is already implemented

Ah, yes. The simulater even wiggles.

> Additional to what we discussed before, as proposed in the last comment, I will:
> 
> - optimize the grid (maybe even bigger),
> - user feedback (haptic feedback is already implemented, but will try to add an additional indicator)
> - add the "blue screen" setting validation
> - and get back to you.

i18n - We'll kick the translations off, its a GitHub task we run, easy for us to initiate.

We also need to consider the documentation updates too. :grimacing:
- Valid PIN range, one for all.
- Timeouts & exiting the app won't evade them
- Amend current warnings about security from "_don't_" to "_at your own risk_".

Great! Thanks for all the engagement. I think there are a few users who are currently using the app insecurely, so this update at least gives them something better.

Just wondering about what to broadcast on https://community.home-assistant.io/t/home-assistant-app-for-garmin/637348. Did you want to promote your change as a contributer once its all merged?

moesterheld:
I addressed some of the comments.
All that's left now is:
- optimizing the input, I am even thinking of a bigger one now after using the calculator apps for a while
- validating the PIN set in settings (after deciding on a grid size)
- update the documentation

The timer should be fine, it works like the confirmation timer, only that it is reset after each touch. When the dialog is left, the timer is destroyed.

I will mark the PR as ready for review again once I am done, no need to check back regularly

moesterheld:
Ok... almost there. 
I have now chosen to go with the full set of digits. I think it performs pretty well and looks good on both round and square displays. I have also added highlighting on tap and some kind of a "progress indicator". Thanks for making me do this, I like it much better 😄

https://github.com/user-attachments/assets/c0964663-5d1a-4787-a5b8-6e92a244872b
round

I didn't want to introduce a breaking change, so right now, the PIN setting is optional. If it is not set, ´"confirm": true´ behaves the way it does right now. If we want to force users to use the pin for every confirmation, we could just set a default PIN. I am not sure how app updates with config changes are handled when using Connect IQ. Do you need to re-set all parameters like in the simulator? I have completely replaced my beta app with a new UID, so I couldn't check that unfortunately.

Also, we could extend the PIN size if we want by simply changing the validation (e.g. to also allow PINs with more than 4 digits). But I don't know if that will confuse the users.

Let me know what you think about the two points, then we can adapt the docs accordingly.

philipabbey:
> If it is not set, `"confirm": true` behaves the way it does right now.

Do you mean `"pin" :true` behaves the same way as `"confirm": true` when the PIN is not set?

This is a significant piece of work. Well done. I'll endeavour to review the changes this week, probably Friday. I'm looking forward to it, and I'll sideload a version onto my watch for a proper test.

Thank you!

philipabbey:
Oh, I liked it the original way, where I could choose either confirm or PIN or neither. I do have menu items presently which are confirm without PIN to avoid accidental initiation but do not need security. Its the difference between one extra click and several. Also, we would be changing the behaviour for existing users, making their confirms more onerous with a PIN.

moesterheld:
No problem, that's easy to change. I'll take care of it tonight

source/HomeAssistantTapMenuItem.mc
```diff
@@ -84,7 +87,20 @@ class HomeAssistantTapMenuItem extends WatchUi.IconMenuItem {
...
 
     function callService() as Void {
-        if (mConfirm) {
+        var hasTouchScreen = System.getDeviceSettings().isTouchScreen;
+        if (mPin && hasTouchScreen) {
+            var pin = Settings.getPin();
+            if (pin.toNumber() == null || pin.length() != 4) {
+                ErrorView.show(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String);
+                return;
+            }
```
philipabbey, line 96: Is it worth moving these check to `Settings.getPin()`? Then we know that wherever `getPin()` is used it returns a valid PIN, and the checks are done centrally and only coded once.

source/Settings.mc
```diff
@@ -164,6 +166,10 @@ class Settings {
         return mConfirmTimeout * 1000; // Convert to milliseconds
     }
 
+    static function getPin() as Lang.String {
+        return mPin;
+    }
```
philipabbey, line 171: **Question:**

Perform conformance checks and return a valid PIN? So I'm wondering if the initialiser would fetch the pin with `Properties.getValue("pin")`, check it conforms to 4 digits [0..9] and store it in `mPin` if it does, otherwise store `null` for invalid PIN.

`private static var mPin as Lang.String or Null = "0000";`

getPin() would then check if `mPin` was `null` and `ErrorView.show(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String);` or return the 4 digit PIN string.

source/HomeAssistantPinConfirmation.mc
```diff
@@ -0,0 +1,307 @@
...
+}
+
+class PinFailures {
+
+    const MAX_FAILURES         as Number = 5;              // maximum number of failed pin confirmation attemps allwed in ...
+    const MAX_FAILURE_MINUTES  as Number = 2;              // ... this number of minutes before pin confirmation is locked for ...
+    const LOCK_TIME_MINUTES    as Number = 10;             // ... this number of minutes
+    
+    const STORAGE_KEY_FAILURES as String = "pin_failures";
+    const STORAGE_KEY_LOCKED   as String = "pin_locked";
```
philipabbey, line 251: Perhaps these belong in `Globals.mc` class with other centrally managed constants?

source/HomeAssistantToggleMenuItem.mc
```diff
@@ -213,19 +216,33 @@ class HomeAssistantToggleMenuItem extends WatchUi.ToggleMenuItem {
...
 
     function callService(b as Lang.Boolean) as Void {
-        if (mConfirm) {
+        var hasTouchScreen = System.getDeviceSettings().isTouchScreen;
+        if (mPin && hasTouchScreen) {
+            var pin = Settings.getPin();
+            if (pin.toNumber() == null || pin.length() != 4) {
+                ErrorView.show(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String);
+                return;
+            }
```
philipabbey, line 225: Is it worth moving these check to Settings.getPin()? Then we know that wherever getPin() is used it returns a valid PIN, and the checks are done centrally and only coded once.

philipabbey (changes requested):
Very well done, but I do have a couple fo suggestions.

Thank you!

philipabbey:
Tests I'm now performing on my Venu 2 watch.
- [ ] Deploy to watch
- [ ] Correct PIN
- [ ] Incorrect PIN multiple times
- [ ] Correct PIN after timeout
- [ ] Change PIN during running
- [ ] Bad PIN, i.e. not 4 digits

moesterheld:
Hi @philipabbey . Thanks for the review. I addressed your comments. 
I moved the validation to `Settings.mc`. I am struggling to show a validation warning on startup, though. The app crashes when moving it into `validatePin()`, not sure why. So right now, the error screen is shown when accessing the value, validation is done beforehand, though.
Maybe I will find the time to take another look later today.

philipabbey:
I'll also try later tonight, to see what the issue is presenting the error. I thought we already had one such error display in the settings, which makes me wonder if it has ever worked...

philipabbey:
Firstly what you have presently represents a suitable solution, failing on `Settings.getPin()`.

```
    static function getPin() as Lang.String or Null {
        if (mPin == null) {
            ErrorView.show(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String);
        }
        return mPin;
    }

    private static function validatePin() as Lang.String or Null {
        var pin = Properties.getValue("pin");
        if (pin.toNumber() == null || pin.length() != 4) {
            // Trying the error message on validation
            ErrorView.show(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String);
            return null;
        }
        return pin;
    }
```

I just tried copying the `ErrorView.show()` across into `Settings.validatePin()` and _argh!_ The simulator just presents a white screen with no error message. And the simulation output window does not give a clue. This looks bad, and very annoying. Perhaps its because it all happens before the main view of the application is shown?

This might not be fixable. I'll have another think to see if we can identify the settings PIN error on startup, but at least we have the current version you are offering to the project.

JosephAbbey (approved):

philipabbey (approved):
> Yes, that is also what i found. But I now found the correct place for it. Works for glance and app view.

Yep, that works.

> [add trailing period to be consistent with other messages](https://github.com/house-of-abbey/GarminHomeAssistant/pull/192/commits/796be5b960338e567724680f2a58b9d30e7e198d)

I think we have been inconsistent here, sorry. I'll pick that up under a "cosmetic" issue/MR, along with a couple of other trivial things.

philipabbey:
Thank you for your considerable contribution.

moesterheld:
@philipabbey @JosephAbbey 
You're very welcome. Thanks for creating this great app!

philipabbey:
Not always a shared view...

image

Well it comes with a 100% money back guarantee.

moesterheld:
😂 but hey, a 4.8🌟 average is worth more than a rating from the one guy who won't rtfm.

PR #249, opened 2025-07-06
247 improve the glance view

philipabbey:
2s seems to be a good compromise between the status flickering and the update delay. When the menu is cached there's no flickering.


philipabbey:
Best to merge this before work on disabling menu items under #250.

PR #252, opened 2025-07-09
243 option to close application after command

### LINKED ISSUE #250: Add an option to menu items to enable and disable them from use
This allows items to be temporarily disabled. For example:
1. Christmas lights only at Christmas
2. Remove summer cooling options in winter.
3. Heating options in summer

This is presently achievable by changing the menu URL. But in doing so, multiple near copies of the menu can become inconsistent with changed Entity IDs and other options.

This option extends to entire menus by disabling the submenu item.

philipabbey:
Also closes #250.


source/HomeAssistantTapMenuItem.mc
```diff
@@ -105,6 +108,9 @@ class HomeAssistantTapMenuItem extends HomeAssistantMenuItem {
         if (mService != null) {
             mHomeAssistantService.call(mService, mData);
         }
+        if (mExit) {
+            System.exit();
```
JosephAbbey, line 112: The service call is not synchronous here, so we could exit before the call has happened and therefore cancel the request unintentionally.

source/HomeAssistantToggleMenuItem.mc
```diff
@@ -294,6 +297,9 @@ class HomeAssistantToggleMenuItem extends WatchUi.ToggleMenuItem {
     //
     function onConfirm(b as Lang.Boolean) as Void {
         setState(b);
+        if (mExit) {
+            System.exit();
```
JosephAbbey, line 301: Set state is asynchronous.

config.schema.json
```diff
@@ -293,6 +320,18 @@
           "required": ["type"]
         }
       ]
+    },
+    "enable": {
```
JosephAbbey, line 324: The standard for HomeAssistant has been to use "enabled": https://www.home-assistant.io/docs/scripts/#disabling-an-action

JosephAbbey (changes requested):

source/HomeAssistantToggleMenuItem.mc
philipabbey: Ah, wrong call back function. That's actually quite annoying as the other one is less convenient to use.

JosephAbbey (approved):

Issue #142, opened 2024-05-15
GHA over Wi-Fi


philipabbey:
@brunoleon suggested it should be able to make the application work over Wi-Fi, see https://community.home-assistant.io/t/home-assistant-app-for-garmin/637348/270. Previously we have looked into this but not found a way. @brunoleon has provided some sample code at https://github.com/brunoleon/garmin-ha-wifi-test.

1. Evaluate the sample code, check it works on own Garmin Venu2.
2. Identify how this solution can be added to the Application (not Widget) version of GarminHomeAssistant.

mortenmathiasen:
Awesome app, but it requires bluetooth connection. What is the status of supporting WiFi?

philipabbey:
Not great. There's been a suggestion it can work, but so far we've not had any success. The "Sync" process over Wi-Fi is for sync'ing (e.g. FIT files), and is not immediately available for general HTTPS, not as far as we have found. This issue is dragging on, and too many people seem to believe because they have Wi-Fi on their watch this must be possible. We're still looking to see if we can hijack the sync process for HTTPS, but we'll need to take the whole process apart and see if we can pretent to be sync'ing. Then we may have to re-architect the App as there are multiple GET requests over Bluetooth that the App needs to work, and the Wi-Fi sync process has an overhead we can't afford more than once.

To be honest, the request is a bit of a nightmare presently

Wi-Fi has additional demands over Bluetooth, e.g. its more power hungry (discharging your watch battery), and I'm sceptical it will deliver the benefit people want and take dispproportionate effort to code.

brunoleon:
I think that it might be wise to be able to explicitely enable wifi when you need it for a feature (I mean an option in the json config you profive to the app)

ex: I need a wifi fallback for opening my garage door, for updating a widget I don't care and prefer to save battery

philipabbey:
Except you need to enable Wi-Fi to get the JSON config (unless cached). Again possible, but new failure modes to code around. The usually off nature of Wi-Fi makes for a quite different proposition. A re-design even.

philipabbey:
I've finally got around to looking at this more closely. I've simplified the code, and removed a class. Here's a video of the transfer of a single file.

https://github.com/user-attachments/assets/81624bed-a044-42ff-8ffa-fd5890746343

[...]

PR #255, opened 2025-07-11
add Wifi LTE command execution

vincentezw:
Firstly, thank you to all the maintainers for your work on this very useful application.

This PR adds the ability to execute Home Assistant commands over Wi-Fi or LTE, behind an opt-in setting. While not a full "Wi-Fi mode," this feature allows the app to function in limited scenarios where the phone is unavailable but the device still has a direct network connection (e.g. Wi-Fi or LTE).

When enabled:
- The app can launch with a cached menu without phone connection.
- On command execution, the user is prompted to confirm.
- Upon confirmation, a bulk sync is triggered to send the request to Home Assistant.

This enables basic control even without Bluetooth connectivity — so I can switch off that pesky bathroom light late at night when I don't have my phone nearby.

I’ve seen a few issues suggesting similar functionality, and I believe this strikes a useful balance between functionality and simplicity. That said, I understand if this doesn't align with the intended featureset — feel free to close if so.

This PR also adds a few string resources without translations. I'm not sure if those are autogenerated or user-provided — happy to adjust if needed.

### DIFF
--- a/resources/settings/properties.xml
+++ b/resources/settings/properties.xml
@@ -96,4 +96,9 @@
   -->
   <property id="webhook_id" type="string"></property>
 
+  <!--
+    Enables the SyncDelegate and prompt to send a command over Wifi/LTE.
+    This will only show when not connected to the user's phone.
+  -->
+  <property id="wifi_lte_execution" type="boolean">false</property>
 </properties>
--- a/resources/settings/settings.xml
+++ b/resources/settings/settings.xml
@@ -116,4 +116,11 @@
   >
     <settingConfig type="alphaNumeric" readonly="true" />
   </setting>
+
+  <setting
+    propertyKey="@Properties.wifi_lte_execution"
+    title="@Strings.WifiLteExecution"
+  >
+    <settingConfig type="boolean" />
+  </setting>
 </settings>
--- a/resources/strings/strings.xml
+++ b/resources/strings/strings.xml
@@ -31,7 +31,9 @@
   <string id="NoInternet">No Internet connection.</string>
   <string id="NoJson">No JSON returned from HTTP request.</string>
   <string id="NoPhone" scope="glance">No Phone connection.</string>
+  <string id="NoPhoneNoCache" scope="glance">No phone connection, no cached menu.</string>
   <string id="NoResponse">No Response, check Internet connection</string>
+  <string id="TimedOut">Request timed out</string>
   <string id="PinInputLocked">PIN input locked for</string>
   <string id="PotentialError">Potential Error</string>
   <string id="Seconds">seconds</string>
@@ -42,6 +44,10 @@
   <string id="UnhandledHttpErr">HTTP request returned error code = </string>
   <string id="WebhookFailed">Failed to register Webhook</string>
   <string id="WrongPin">Wrong PIN</string>
+  <string id="WifiLteNotAvailable">No Wifi or LTE available</string>
+  <string id="WifiLtePrompt">Execute over Wifi/LTE?</string>
+  <string id="WifiLteExecutionTitle">Sending to Home Assistant.</string>
+  <string id="WifiLteExecutionDataError">No data received.</string>
 
   <!-- For the settings GUI, strings should be in the order they are used. -->
   <string id="SettingsSelect">Select...</string>
@@ -64,4 +70,5 @@
   <string id="SettingsEnableBatteryLevel">Enable the background service to send the device battery level, location and (if supported) activity data to Home Assistant.</string>
   <string id="SettingsBatteryLevelRefreshRate">The refresh rate (in minutes) at which the background service should repeat sending data.</string>
   <string id="WebhookId">(Read only) The Webhook ID created by the device for background service updates. You might require this for debugging.</string>
+  <string id="WifiLteExecution">Allows the app to start without phone connection, and prompt to execute command over Wifi/LTE.</string>
 </strings>
--- a/source/HomeAssistantApp.mc
+++ b/source/HomeAssistantApp.mc
@@ -14,6 +14,7 @@
 //-----------------------------------------------------------------------------------
 
 using Toybox.Application;
+using Toybox.Communications;
 using Toybox.Lang;
 using Toybox.WatchUi;
 using Toybox.System;
@@ -122,11 +123,14 @@ class HomeAssistantApp extends Application.AppBase {
         } else if (Settings.getPin() == null) {
             // System.println("HomeAssistantApp getInitialView(): Invalid PIN in application settings.");
             return ErrorView.create(WatchUi.loadResource($.Rez.Strings.SettingsPinError) as Lang.String);
-        } else if (! System.getDeviceSettings().phoneConnected) {
-            // System.println("HomeAssistantApp getInitialView(): No Phone connection, skipping API call.");
+        } else if (! System.getDeviceSettings().phoneConnected and Settings.getWifiLteExecutionEnabled() and ! hasCachedMenu()) {
+            // System.println("HomeAssistantApp getInitialView(): No Phone connection, no cached menu, skipping API call.");
+            return ErrorView.create(WatchUi.loadResource($.Rez.Strings.NoPhoneNoCache) as Lang.String);
+        } else if (! System.getDeviceSettings().phoneConnected and ! Settings.getWifiLteExecutionEnabled()) {
+            // System.println("HomeAssistantApp getInitialView(): No Phone connection and wifi disabled, skipping API call.");
             return ErrorView.create(WatchUi.loadResource($.Rez.Strings.NoPhone) as Lang.String);
-        } else if (! System.getDeviceSettings().connectionAvailable) {
-            // System.println("HomeAssistantApp getInitialView(): No Internet connection, skipping API call.");
+        } else if (! System.getDeviceSettings().connectionAvailable and ! Settings.getWifiLteExecutionEnabled()) {
+            // System.println("HomeAssistantApp getInitialView(): No Internet connection and wifi disabled, skipping API call.");
             return ErrorView.create(WatchUi.loadResource($.Rez.Strings.NoInternet) as Lang.String);
         } else {
             var isCached = fetchMenuConfig();
@@ -227,6 +231,20 @@ class HomeAssistantApp extends Application.AppBase {
         WatchUi.requestUpdate();
     }
 
+    //! Can we use the cached menu?
+    //!
+    //! @return Return true if there's a menu in cache, and if the user has enabled the cache and 
+    //! has not requested to have the cache busted.
+    //
+    function hasCachedMenu() as Lang.Boolean {
+        if (Settings.getClearCache() || !Settings.getCacheConfig()) {
+            return false;
+        }
+
+        var menu = Storage.getValue("menu") as Lang.Dictionary;
+        return menu != null;
+    }
+
     //! Fetch the menu configuration over HTTPS, which might be locally cached.
     //!
     //! @return Return true if the menu came from the cache, otherwise false. This is because fetching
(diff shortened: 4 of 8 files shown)


philipabbey:
Hello! I will be very grateful if you are able to pull this off. We have had a few issues with getting Wi-Fi to work with this app. We can get an App to send an API request over Wi-Fi. But I can't get it to send an API request from a menu item. I can get it work in simulation, but then on the watch it managed to crash the device. I had given up in dispair as I had run out of ideas. I assumed the problem was running the code off different threads of execution. Essentially, I think the sync function was not provided with a more general purpose API GET/POST request usage in mind. This was where I got to: #142.

I can assist by testing your code on my Venu 2 watch. There are others out there you might be able to enlist, e.g. @brunoleon, @mortenmathiasen and @martin-snajdr may be willing to offer some testing time to you.

Good luck!

Philip

philipabbey:
> This PR also adds a few string resources without translations.

Don't worry about those, we have a Python script for automated translation that includes user supplied manual overrides for when the translations don't come out well. We can fix the translations after we've got working code, there's a build task to set off for those.

vincentezw:
Hi, thanks for the quick reply! For context, I've been testing this with success on my Venu 3. I've used the same pattern in a different app for a little longer, where it works well.

Would love to hear from others with different devices, and thanks for giving it a go on your Venu 2.

In general I think the PR may need to wrap some calls in `has` conditions a bit more dilligently, for instance `sync2` was introduced in API 4.0.4. But it should be easy to gate it from `Settings.getWifiLteExecutionEnabled`.

philipabbey:
I've compile locally and published to a beta app for testing this afternoon.

I can't find any `sync2` reference, only [`SyncDelegate`](https://developer.garmin.com/connect-iq/api-docs/Toybox/Communications/SyncDelegate.html) which you appear to have used. That appears to be available at API Level 3.1.0 and the `manifest.xml` specifies `minApiLevel="3.1.0"`, so that's okay right?

vincentezw:
It's [this instance](https://developer.garmin.com/connect-iq/api-docs/Toybox/Communications.html#startSync2-instance_function) method I'm using which is available since 4.0.4.
But I should add a conditional so we can use [startSync](https://developer.garmin.com/connect-iq/api-docs/Toybox/Communications.html#startSync-instance_function) (api level 3.1.0) when `startSync2` isn't available.

philipabbey:
Ah `startSync2`, I'm following now. Sadly I got a black screen on starting your version of the App. :-(

https://developer.garmin.com/connect-iq/compatible-devices/ suggests Venu 2 watched should be on API 5.0. So your choice of call should not affect me.

Okay, it worming now... I'll have to see if that glitch is repeatable. I've turned on your new option to use Wifi/LTE and turned off my phone's Wi-Fi and Mobile data. The watch then complains "No Response, check Internet connection" with a blue screen. It does not offer to use the watch's Wi-Fi.

What am I missing here?

JosephAbbey:
> > > > Ah `startSync2`, I'm following now. Sadly I got a black screen on starting your version of the App. :-(
> > > > https://developer.garmin.com/connect-iq/compatible-devices/ suggests Venu 2 watched should be on API 5.0. So your choice of call should not affect me.
> > > > Okay, it working now... I'll have to see if that glitch is repeatable. I've turned on your new option to use Wifi/LTE and turned off my phone's Wi-Fi and Mobile data. The watch then complains "No Response, check Internet connection" with a blue screen. It does not offer to use the watch's Wi-Fi.
> > > > What am I missing here?
> > > 
> > > 
> > > Is caching enabled and the menu cached? It should allow you to start the app without that error so you can execute a command.
> > 
> > Am I understanding correctly that the menu is not fetched by WiFi or LTE? Is that what we want? I'm curious what others think.
> 
> Yeah, this intentionally only works with a cached menu, originally retrieved over BLE. 
> 
> It would be possible to retrieve the menu over Wifi/Lte but that would mean an initial (rather lengthy) sync before even being able to use the app, with another sync progress to execute a command. I did not deem that worthwhile but happy to take on feedback.
> 
> A further improvement could/should be to retrieve the menu *after* sending a command with a second request, since we can simply loop requests together in the delegate. If we're on Wifi, we may as well make that extra call. But that still means using a cached menu to execute a command to begin with.

My only thought was for someone who doesn't have their phone on them, hence LTE or WiFi, and doesn't have the menu cached may need to download it. Obviously if the menu is cached then use it, but in the case it isn't, I think we should download it, maybe with a popup to confirm?

philipabbey:
If the menu is not cached we ought to disable the Wi-Fi option.

vincentezw:
> My only thought was for someone who doesn't have their phone on them, hence LTE or WiFi, and doesn't have the menu cached may need to download it. Obviously if the menu is cached then use it, but in the case it isn't, I think we should download it, maybe with a popup to confirm?

That sounds like a nice enhancement. I would prefer to implement something like that as a future improvement, rather than in this PR though. Similar to the final pull after sending a command.

I do think we need to acknowledge that with existing APIs CiQ offers, the Wifi/LTE mode is clunky and more a "last resort" than a viable operating mode. The sync needs to be triggered, consumes battery, and takes time.

philipabbey:
> I do think we need to acknowledge that with existing APIs CiQ offers, the Wifi/LTE mode is clunky and more a "last resort" than a viable operating mode.

Very much agree. I've spent too much time fielding people who say "but my watch has Wi-Fi". "Why can't you just...?"

philipabbey:
> A further improvement could/should be to retrieve the menu after sending a command with a second request

If we did, it might be easiest for that menu to only come into being the next time the app is started. Otherwise we need to ditch the constructed menu and reconstruct. Perhaps rebuilding the menu is not so bad? But we have to remember its just a watch app for doing a few things one can't be bothered to reach for one's phone for. So I'm loathed to bloat the software and cause it to use oodles of memory.

I'm quite happy that the menu must be cached TBH. But the simplification does mean making sure users can't get the app into a non-working state. They then complain and bad mouth the app on the reviews, which I feel is a bit rich and lazy for free software!

So if we keep that simplification, I would prefer to keep the complaints at bay by preventing them, That means disabling the Wi-Fi option unless the menu is already cached. But we're a bit away from that level of detail presently! (https://developer.garmin.com/connect-iq/core-topics/properties-and-app-settings/ indicates it should be possible.)

vincentezw:
Oh neat, I get what you mean now. The setting to enable the Wifi/Lte stuff should not be allowed to be enabled, when menu caching is not enabled.

Will take a look at that. If not feasible, we could always have `Settings.getWifiLteExecutionEnabled` return `false` is the caching isn't enabled, maybe with a toast or something.

philipabbey:
> The app should already refuse to start without a phone connection, when there's no cached menu and the Wifi feature is enabled

We can cover that in the documentation, if people actually read it properly... Hence ideally making limitations very obvious.

philipabbey:
It's no fluke either, but....! The changes made have managed to bypass the 'confirm' and 'PIN' options. I managed to open our garage doors over W-Fi without entering the PIN. So that's one to watch! Just need to check through the conditions and not bypass the (rather basic) security!

vincentezw:
There is an issue with the `popView` the `HomeAssistantPinConfirmationDelegate` does. It seems to behave a little odd. It's responsible for closing itself when the correct pin is entered, but it ends up closing the next dialog (the wifi confirmation) when doing the Wifi sync instead of itself.
But if I remove that `popView` it doesn't close itself on the normal BLE route.
Will take a closer look in a while. It's hard to keep track of the state of the toggles, the way we're forced to pass that state in so many places 😅

vincentezw:
I added a number of improvements in the three commits above. In 8b55680a5091000291b66891f6108d35138d8476 I flip the toggles for empty response bodies, which should fix https://github.com/house-of-abbey/GarminHomeAssistant/issues/256 in the most straightforward way I can think of.

philipabbey:
I'll be able to compile your latest version later next week. Sorry for the delay. Looking forward to it.

Philip

brunoleon:
Happily discovering this was not dead.
I can give it a go next week too I think, on an Epix Pro gen 2.

vincentezw:
I had a quick test again here on my side, but things are working correctly for me thus far. I have an entry like this:
```
        {
          "entity": "light.zoe_bedside",
          "name": "Zoe light tog",
          "type": "toggle",
	  "tap_action": {
	    "pin": true
	  }
        }
```
and am prompted for the PIN as expected followed by the Wifi confirmation. It works fine in the simulator with a Venu 2 too. Am I missing anything in my replication steps?

 Good question re. https://github.com/vincentezw/GarminHomeAssistant/blob/main/source/HomeAssistantApp.mc#L589 - I vaguely remember not quite following why certain `ErrorView.show` calls wouldn't do anything for me, so will take a closer look at that.

vincentezw:
FIWI I'm not a VScode user. Neovim doesn't like the hoover documentation in Monkey C at all sadly. I think it may be expected that a nested list doesn't render correctly, unless that works for you elsewhere.

philipabbey:
Bummer!

Which SDK version have you chosen?

It's not working for me in simulation or on the Venu 2 watch.

I'm having feelings of deja vu. These feel like race conditions from the last experiment.

I'm looking to see what changes my end might make the code work,  and I'll report back.

vincentezw:
Running on 8.2.1 on intel Mac.

I added an early return from `fetchApiStatus`, when in the app and on wifi. I don't really think that caused issues though, or at least nothing introduced by this PR. The pattern in `getInitialView` is a bit odd, we call `fetchApiStatus` but don't return the `ErrorView` it shows but rather the cached menu.

philipabbey:
> The pattern in `getInitialView` is a bit odd, we call `fetchApiStatus` but don't return the `ErrorView` it shows but rather the cached menu.

The purpose of the `fetchApiStatus()` is to test the API is available and set the status via the HTTPS request's call back. I see your point about returning the `ErrorView`s. These views would not constitute the "initial" view, but a dismissable view on top of the inital view. I have not tested what happens if the `ErrorView`s become the initial view, and if dismissing them would exit the App. Presently dismissing them should reveal the intended initial view.

vincentezw:
I don't think it would explain your issues, but I do think we need to add an `onBack` to the `HomeAssistantPinConfirmationDelegate`. Playing around with it on my watch, going back and forth a few times becomes a bit glitchy. I think we don't stop the timer, which then ends up popping whatever view we're in. And it also needs to untoggle the toggle again. Will take a look at that later today.

For full disclosure; using the Wifi mode exposed a number of oddities with the implementation, which is why I added the callbacks in 95b5cbc0e5e6ba52f3b44542f5f72eb6d4ecd218. These bugs are masked in the "normal" non-wifi mode when the callback from the request will "fix" the inconsistent toggle state after a few moments.

vincentezw:
Actually, the "confirm" dialog doesn't seem to work correctly for me anymore on wifi either (the pin one does). Will take a peek at that later today, probably a small oversight, as I'm pretty confident this was okay when I ran it last week.

vincentezw:
> > The pattern in `getInitialView` is a bit odd, we call `fetchApiStatus` but don't return the `ErrorView` it shows but rather the cached menu.
> 
> The purpose of the `fetchApiStatus()` is to test the API is available and set the status via the HTTPS request's call back. Neither this nor the callback return the menu. That's the job of `fetchMenuConfig()` which returns a boolean indicating if the menu is cached. the callback sets the menu via side effect (assigns a field).
> 
> I see your point about returning the `ErrorView`s. These views would not constitute the "initial" view, but a dismissable view on top of the inital view. I have not tested what happens if the `ErrorView`s become the initial view, and if dismissing them would exit the App. Presently dismissing them should reveal the intended initial view.

I haven't tested it fully, but I suspect those `ErrorViews` are simply not showing, as they'll be popped by the cached MenuView almost immediately.

philipabbey:
I agree with your logic. In my case it was simply that the existing confirmation blocked me from proceeding. I'm up for your suggestions here. But hey, this could all end up working well!

vincentezw:
> I agree with your logic. In my case it was simply that the existing confirmation blocked me from proceeding. I'm up for your suggestions here. But hey, this could all end up working well!

I can imagine that concurrent timers popping views can leave things in a weird state! Alternative to some of the above may be to use a different view for our yes/no screens, so we can use an `InputDelegate` and handle the back button.

vincentezw:
you will want to remove these secrets accidentally commited I think https://github.com/house-of-abbey/GarminHomeAssistant/blob/72059fbd8fd72c02392cc5a2023e5130b1b879b6/resources/settings/properties.xml#L16

vincentezw:
will take a look this evening. Curse those toggles 😅

philipabbey:
> We treat 0 as an explicit instruction to disable polling entirely.

image

We've already defined `0` as no additional delay in the settings. I'm loathed to change that behaviour to 'disabled' now as that would be problematic.

Thanks for the other fixes.

vincentezw:
Gotcha — I hadn’t realised the setting was explicitly defined that way.

That said, I don’t think immediately calling `updateMenuItems` makes sense — not just from a practical standpoint, but also because tight polling loops are generally bad practice. I think a sane fallback delay (e.g. 2 seconds) would be a safer default, or alternatively a separate setting to explicitly disable polling if needed.

Happy to go with whatever makes the most sense, but calling immediately feels risky with little benefit.

philipabbey:
Okay that last commit seems entirely sensible. Let the transfer complete, wait 2s before resuming.

I'll deploy a local test app tomorrow evening. Hopefully that'll cover it for the pull request. Well done! A rather involved job.

I've a question about `Lang.Dictionary`. I prefer your syntax `options[:key]`. Is it entirely 1:1 compatible with `options.get(:key)`? I'm looking to code tidy the other files to be consistent with the square brackets syntax.

vincentezw:
We can always add a setting later to disable the polling if needed. Thanks for all your testing!

Afaik the difference between those two methods of accessing the value is just stylistic. I had thought that the `option[:key]` syntax was a direct access, without the safety check, but in my testing it returns `null` all the same for non-existent keys.

philipabbey (approved):
A significant improvement to the functionality of the application. Thank you!

PR #262, opened 2025-07-24
259 code tidy

philipabbey:
Should be functionally equivalent.


JosephAbbey (approved):

vincentezw (approved):
LGTM! Not necessarily here, but we could also consider using `import` rather than `using` to load classes into the namespace. This would mean we can skip the repetitive `Lang.Array`, `Lang.Boolean` etc. and just use `Array`, `Boolean` etc. See: [the import statement](https://developer.garmin.com/connect-iq/monkey-c/monkey-types/#:~:text=println(%22globalX%20%3D%20%22%20%2B%20globalX)%3B%0A%7D-,The%20import%20Statement,-In%20traditional%20Monkey)

philipabbey:
We seem to have the usual differences in translation between a local Python script run and one run on the server as an action. Since the server is not obliging us presently with the script failing after a partial run with a lack of translation strings, we are forced to use the locally compiled translations and hence we are churning the translated text. Just a note since it might be observed by existing users who are wondering why.

Thanks for your reviews.

PR #266, opened 2025-08-08
Add TLS 1.2 compatibility guidance for Garmin Wi-Fi/LTE connection issues

aalaei:
This update adds a note to the “Limits of Use” section explaining that some Garmin devices use TLS 1.2 for HTTPS handshakes. If a server or proxy enforces a minimum TLS version of 1.3 or higher, users may experience SSL handshake errors with the message HTTP request returned error code = 0.

The documentation now includes:

An explanation of the issue

How to identify the error

A practical solution example for Cloudflare Tunneling users to lower the minimum TLS version to 1.2

A security note advising against lowering TLS below 1.2

This helps users troubleshoot connectivity problems when using Wi-Fi or LTE features on Garmin watches with Home Assistant.

### DIFF
--- a/Wi-Fi.md
+++ b/Wi-Fi.md
@@ -16,6 +16,11 @@ With version 3.0 onwards the application now includes the ability to temporarily
 
 4. Remember that you need to be within range of your watch's configured Wi-Fi access point to utilize this functionality. If supported by your device, LTE offers a longer range, but network charges may apply.
 
+5. On some Garmin devices, the HTTPS handshake is performed using **TLS 1.2**. If your server or proxy enforces a higher minimum (e.g., TLS 1.3), you will encounter an SSL handshake error with the message: `HTTP request returned error code = 0`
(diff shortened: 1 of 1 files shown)


aalaei:
In my setup, I had the minimum TLS version set to 1.3. Over Bluetooth, there were no issues because the watch communicates with the companion smartphone, and the TLS 1.3 handshake is performed by the phone, which supports it without any problem.

When I saw the new Wi-Fi/LTE standalone feature in this Garmin app and actually used it, I encountered the handshake error. After some investigation, I discovered that Garmin devices themselves still use TLS 1.2 and cannot complete a TLS 1.3 handshake.

I thought adding this note could help others who run into the same situation avoid spending hours troubleshooting, as it took me some time to track down the cause.

philipabbey:
Thank you so much for this detailed technical insight. I wonder if you would mind adding your point about the Bluetooth being unaffected as it's solved by the phone, hence a Wi-Fi only issue? Then I'll be merging the revised text.

philipabbey:
Excellent, and thanks for your help and expertise with this unforeseen consequence!

Issue #264, opened 2025-08-07
Add custom HTTPS Header(s) for e.g. Cloudflare Tunnels


tispokes:
Hi, my instance is behind a Cloudflare Tunnel (WAF) and secured to a certificate AND special for my Garmin watch with a custom HTTPS Header. The other Garmin app ( https://github.com/hatl/hasscontrol --- https://apps.garmin.com/apps/3dce2242-473f-4f13-a6a9-299c3686611f ) has this feature, because Garmin does not allow self signed certificates.
But your app has more features as hasscontrol does, may you could add this 2 field to the settings and to the GET/POST in your code?

Like this

<img width="1145" height="2048" alt="Image" src="https://github.com/user-attachments/assets/7845eaa9-4761-4b3b-a5ce-7f626b889bc1" />

tispokes:
> I think you are asking if we can include a custom HTTPS header with a user definable key and user definable value.

Exactly


> E.g. I build a prerelease version of the app for your watch and you test the functionality for me?

Sure no problem, I think. Just have to install somehow the alpha version. Or a branch, how ever the deployment on Garmin watches is working. Because I use it with the other Garmin app, I know it is working

Yeah world be really great. Think I can add some description too (links to cloudflare addon ...)

Lars

philipabbey:
I would really appreciate some help with the documentation here, the why and how one knows what to set up. I'll see if I can start this weekend.

philipabbey:
Much as I had imagined, but I'm reluctant to be seen blatantly copying another app! The author will have worked hard for his product.

philipabbey:
> Currently overlooking the german translation/correction, any naming convenice on branch names?

I would welcome improvements to the translation strings above what automated translation can achieve. As far as branch names go, "beggers can't be choosers", so I'll take what you can offer us.

Thanks,

Philip

philipabbey:
So in short, for those using the Home Assistant add-on [Cloudflared](https://github.com/brenner-tobias/addon-cloudflared), they need to set up an additional HTTP header of their choice to make this application work with their HA instance.

Thank you.

tispokes:
Yes or some other kind of reverse proxy or something

philipabbey:
Oh dear, its been such a long time since I set this up I could not remember the mechanism. I think the need for a developer key means you would have to compile locally yourself to sign the `.iq` file.
* Developer Key: https://developer.garmin.com/connect-iq/core-topics/security/
[...]

PR #268, opened 2025-08-09
Fix translation

tispokes:
Hi, me again

1) I think I had to update the german strings.xml as well (via translation.py) or do you do this before deployment?
2) My idea of getting rid of the double meaning with the PIN, see english strings.xml :-)
3) fixed depicated findAll and tested it, obisously


philipabbey:
See https://github.com/house-of-abbey/GarminHomeAssistant?tab=readme-ov-file#submitting-corrections-for-translations.

Please do not amend `strings.xml`. As you will see from my latest commit on #264, your offered corrections in `corrections.xml` have been incorporated into `strings.xml` by the `translate.py` script. This script uses `corrections.xml` first then automatic translation. Thus any changes to `strings.xml` that you make will be overwritten.

We now have a race between changes on #264 and this PR. Let's untangle.

Accept:
1. `corrections.xml` with the caveat at https://github.com/house-of-abbey/GarminHomeAssistant/pull/265#discussion_r2263636115. I made these two string changes on #264 to expedite them.

* https://github.com/house-of-abbey/GarminHomeAssistant/blob/264-add-custom-https-headers-for-eg-cloudflare-tunnels/resources/strings/strings.xml#L67
* https://github.com/house-of-abbey/GarminHomeAssistant/blob/264-add-custom-https-headers-for-eg-cloudflare-tunnels/resources-deu/strings/corrections.xml#L75

This is the change from "confirmation" to "it" to remove ambiguity with PIN vs confirm fields in JSON.

I also re-ordered the strings to match the settings dialogue order (my previous oversight not yours).

https://github.com/house-of-abbey/GarminHomeAssistant/blob/264-add-custom-https-headers-for-eg-cloudflare-tunnels/resources-deu/strings/corrections.xml#L68-L70

2. `translate.py` - Thanks.

Revert 
1. `**/strings.xml` - No changes to these files.

Sorry, I thought we were done!

tispokes:
Yeah same thought, because all are minor fixes. But discussed about the NoConfigUrl translation by lunch. So I would suggest this:

```monkey-c
  <string id="NoAPIKey" scope="glance">Kein API-Schlüssel in den App-Einstellungen hinterlegt.</string>
  <string id="NoApiUrl" scope="glance">Keine API-URL in den App-Einstellungen hinterlegt.</string>
  <string id="NoConfigUrl" scope="glance">Keine Menükonfigurations-URL (JSON) in den App-Einstellungen hinterlegt.</string>
``` 

And also this fix, I have overseen (remove the -)
`  <string id="WifiLteExecution">Ausführung über WLAN/LTE einschalten.</string>`

tispokes:
And correction for the new header

My simulation is not running so I can't see the menu, where do you use `SettingsUserHttpHeader`
```monkey-c
  <string id="SettingsUserHttpHeader">Zusätzlicher HTTP-Header</string>
  <string id="SettingsUserHttpHeaderDescription">Einige Home Assistant-Installationen erfordern die Angabe eines benutzerdefinierten HTTP-Headers, um zu funktionieren.</string>
  <string id="SettingsUserHttpHeaderKey">Zusätzlicher HTTP-Header: Key</string>
  <string id="SettingsUserHttpHeaderValue">Zusätzlicher HTTP-Header: Value</string>
  ```

PR #272, opened 2025-08-10
264 add custom https headers for eg cloudflare tunnels

philipabbey:
Additional settings to provide a custom HTTP header so that the application can work with Cloudflare's Web Application Firewall (WAF).


philipabbey:
There remains an app crash when opening the app from the glance before the API check goes green. 

This exists in ve r 3.0 and possibly before too.

.vscode/settings.json
```diff
@@ -4,6 +4,8 @@
     "Venu"
   ],
   "files.exclude": {
-    "resources-*": true
+    "resources-*": true,
+    "bin": true,
+    "export": true
   }
 }
```
tispokes, line 11: New empty line missing

tispokes:
Maybe something like a timeout or "is updating" jump into a waiting screen, which you could exit, and then turn into the actual json when loaded. But don't know how you set it up now, may can check later

tispokes:
Another step-by-step which is maybe worthy to mention: https://kcore.org/2024/06/28/using-cloudflare-zerotrust-and-mtls-with-home-assistant-via-the-internet/

.vscode/settings.json
philipabbey: Done. I've never been particularly bothered about new lines at end of files.

philipabbey:
> Another step-by-step which is maybe worthy to mention: https://kcore.org/2024/06/28/using-cloudflare-zerotrust-and-mtls-with-home-assistant-via-the-internet/

That looks excellent.

.vscode/settings.json
tispokes: My old senior dev was obsessed with linting and new lines 😅

philipabbey: Worth knowing. I try to avoid marking an issue as resolved if I did not raise it.

tispokes: Ahh the issue you meant. Sure

philipabbey:
I can't recreate this IQ crash in simulation.

I think it might have been introduced by the "improvement" on the glance view. Perhaps related to the template evaluation, and a race condition between exiting the glance to start the app, and running the call back function for an API request.

I should be able to test that by turning off the glance template. Tomorrow.

philipabbey:
Additional code changes co-authored with reviewer, so review complete.

PR #287, opened 2025-09-13
Update the translation script to use Gemini instead of Google Translate

### LINKED ISSUE #286: Autogenerated translations are bad
The autogenerated Polish translations are so bad, that I'd almost prefer the app to not be translated at all. For example, `unavailable` is translated as `not fluid`, `wrong PIN` is translated as `bad needle`, the name of `Home Assistant` was translated where the official name never is, as well as many minor syntax things. Initially I intended to provide the Polish translations myself, but then I realized that this is probably a more general issue in other languages as well.

Have you considered using modern AI instead of Google Translate for translations? I tried it out, and it does a MUCH better job than Google Translate, since a lot of context is required to translate these strings correctly. The only problem is getting an API key for those, which couldn't be commited to the repo, but e.g. Gemini 2.5 Flash is free and just requires a few clicks. For best results, the script would likely also need to be modified to provide the AI with the entire file, rather than singular strings, so that it can get more context about what it's doing and keep naming consistent throughout the translation.

Or, alternatively, Google Translate also seems to do a slightly better job if you give it the entire file as input for more context, but I feel like trying to get Google Translate to output parsable XML is going to be a nightmare, and I still see some issues in the translations even if that could be managed.

krzys-h:
This pull request rewrites the automatic translation script to use Gemini instead of Google Translate. This results in a huge improvement in the accuracy of translations.

The use of Gemini requires an API key from Google, which you can get at https://aistudio.google.com/app/apikey. The `gemini-2.5-flash` model we are using is free for up to 250 requests per day. We send one request per language (per language, not per translated string!).

The new translations are better in the following ways:
* The AI is explicitly told that it's working on a smartwatch app, and that shorter, UI-friendly translations are preferrable
* Instead of asking for a translation of a single string, the AI agent gets all the English strings, all the already translated strings, and all the strings to translate at once. This helps it keep a consistent translation style across all strings in the file.
* It also has access to the string keys, which sometimes give a context clue about how to translate something properly. For example, the string "Sure?" was translated wrong across many languages, because this single word with a question mark by itself doesn't really suggest what it's supposed to be.
* It has an explicit rule to never translate app names such as Home Assistant
* When translating comments, it prefers to keep the already translated string if the English meaning didn't change, instead of regenerating the translation on every single run

To improve all of the existing translations without regenerating everything scratch, I also added an "improve mode" option to the script. Improve mode significantly changes the behavior - instead of only translating new strings, the AI is allowed to change whatever it feels suitable, under the condition that the change corrects translation mistakes, grammatical errors or style. Otherwise, the old translation is kept.

I can't verify any languages other than Polish, but the Polish translation is significantly improved after these changes.

The code of the script itself was also written with help of AI. `corrections.xml` is still authoritative in all cases.

You could likely get away with providing the AI with both xml files and asking it to do what it needs to do itself, but keeping the manual parsing gives me more confidence that the output will be correct. That would simplify a lot of this code, though.

Closes #286


JosephAbbey:
Not doing a full code review yet.

We will need to change the GitHub action, to make sure that the API key gets included.
We can set the API key in the repo settings.

When we merge this request in, we will discard the generated strings files, and once all strings have "Home Assistant" updated to "HomeAssistant" (#288), we will run the GitHub action to translate.

Issue #284, opened 2025-09-11
v3.2 -> v3.3 much less menu items possible


Cub3r4:
**Describe the bug**
I used v3.2 without problems for a long time. Today I updated to 3.3. First, let me say thanks for the auto-update feature! It is fantastic! Nevertheless, my first impression after the update was: nothing works: iQ!-error on my INSTINCT 2X TACTICAL. After reducing the amount of items in the garmin.json, everything works fine again.
But sadly, I had to remove a lot. Before, 41 were possible. Now, with v3.3 my limit is 16! Had to cut down the json alot!
If that was nothing intentional from you guys (like a sacrifice for implementing sth else) maybe you can bring back my missing entities :') or even more! ;)

If I add only one more item to my now 16 -> IQ!-error right after auto-resyncing with the garmin.json.

I absolutely did nothing else on my garmin watch, than g to connect store and hit the update button for this app. before 41 max, now 16 max. Everything else works just as fine as before.

**IMPORTANT!**
I have:

- [ ] Used HTTPS to access HomeAssistant (with a public certificate) no, but I use the workaround with a garmin http adress and rerouting to hass. worked perfectly ever since and also does now.
- [x ] Checked that my JSON Syntax is correct ([options 2, 3, or 4](https://github.com/house-of-abbey/GarminHomeAssistant#editing-the-json-file) or otherwise)
- [x ] Checked that my JSON follows the schema ([options 2, 3, or 4](https://github.com/house-of-abbey/GarminHomeAssistant#editing-the-json-file))
- [x ] Used the online editor to check my JSON against HomeAssistant ([option 1](https://github.com/house-of-abbey/GarminHomeAssistant#editing-the-json-file))
- [x ] Checked the [Docs](https://github.com/house-of-abbey/GarminHomeAssistant)
- [x ] Checked the [Troubleshooting Guide](https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/TroubleShooting.md)

**Expected behavior**
should be clear

**Garmin Device (please complete the following information):**

- Model: Instinct 2X Tactical

**Smartphone (please complete the following information):**

- Device: Xiaomi Mi 9T Pro
- OS: Android 15

**Additional context**

philipabbey:
[...]

PR #298, opened 2025-10-13
Add Numeric Menu Item

thmichel:
Added a new numeric menu item to set numeric values e.g. for heating, volume, dimmer etc.

### DIFF
--- a/source/HomeAssistantNumericMenuItem.mc
+++ b/source/HomeAssistantNumericMenuItem.mc
@@ -177,13 +177,11 @@ class HomeAssistantNumericMenuItem extends HomeAssistantMenuItem {
     //
     function getNumericTemplate() as Lang.String? {
         var entity_id = mData["entity_id"];
-        var attribute = mPicker["attribute"] as Lang.String;
+        var attribute = mPicker["attribute"] as Lang.String?;
         if (entity_id == null) {
             return null;
         } else {
             if (attribute == null) {
-                // Compiler says: "Statement is not reachable."
-                // This is wrong because a break point on the following line proves it is executed!
                 return "{{states('" + entity_id.toString() + "')}}";
             } else {
                 return "{{state_attr('" + entity_id.toString() + "','" + attribute + "')}}";


philipabbey:
This is an interesting idea. Firstly let me warn you about a recent issue we had. https://github.com/house-of-abbey/GarminHomeAssistant/issues/284#issuecomment-3284966418.

> So in short, I am starting to reject clever ideas for increased functionality in order to keep the app working on older devices.

This issue was caused by adding more features, leaving some older devices with insufficient memory to cope. So yes your idea does sound like fun, and my watch would probably cope with the increased memory load, but I feel sure other watches may notice the added functionality in an adverse way. Worse, it is not possible to package up the new application for pre-release testing with selected users outside of the app store. Sideloading does not fully work as the options are not settable. So this means we don't get the chance to check the memory drain before release. All this makes me nervous about adding new functionality from now on.

I have now measured the additional memory usage of your PR #298 over the current main branch on an `instinct2x` device.

Total memory available apps in the simulator is 98,304 bytes according to https://developer.garmin.com/connect-iq/device-reference/instinct2x/. When I measure than in the simulator (only) its slightly less at  94,112 bytes.

Current Main:
Used: 62,360 
Free: 31,752 = 33.7%

PR #298:
Used: 66,008 
Free: 28,104 = 29.9%

So the code uses 3.9% extra memory or 3,648 bytes. Does not sound like a lot, but its 11.5% of the free working memory needed to work with the menu definition.

For newer devices, this won't be a problem, e.g. Venu2 has  781,976 bytes for apps.

The bottom line is I like your idea, but I think we'll get complaints about further reductions to menu size on older devices. I do not think there is a way to manage conditional inclusion of features based on device model. And if we did we would have to list two separate apps in the ConnectIQ store. And I don't want the complication in the code base.

Options:
1. Drop old devices like `instinct2x` and allow the code base to move forwards. (Aside: I've had issuees with icons on this device which I have given up trying to solve.)
2. Reject the PR (for now), which is a real pity. Its a pretty basic idea for functionality that is cheaper than say the PIN idea.

I'll have to mull on this. In the meantime I am surveying all devices for memory requirements. I might have found a way to extract the right field from each device in the SDK. That way we can see how many devices are affected.

FYI @Cub3r4

source/HomeAssistantNumericMenuItem.mc
```diff
@@ -0,0 +1,231 @@
...
+
+    ///! Increase value when Up button is pressed or touch screen swipe down 
+
+    function increaseValue() as Void {
+        if (mValueChanged)
+        {
+            mValue += mStep;
+        }
+        else {
+            mValue= getSubLabel().toFloat() + mStep;
```
JosephAbbey: Should use `mValue` not `getSubLabel().toFloat()` here as well. (See comments on `updateState`).

source/HomeAssistantNumericMenuItem.mc
```diff
@@ -0,0 +1,231 @@
...
+    //!
+    //! @param data The rendered template (typically a string) to be placed in the sub label. This may
+    //!             unusually be a number if the SDK interprets the JSON returned by Home Assistant as such.
+    //
+    function updateState(data as Lang.String or Lang.Dictionary or Lang.Number or Lang.Float or Null) as Void {
+        // If vlue has changed, don't use value from HomeAssitant but display target value
+        if (mValueChanged) {
+            setSubLabel(mValue.format(mFormatString));
+            WatchUi.requestUpdate();
+            return;
```
JosephAbbey: This early return will prevent `mValue` updating if `mValueChanged`.

source/HomeAssistantNumericMenuItem.mc
```diff
@@ -0,0 +1,231 @@
...
+    }
+    
+
+    //! Callback function after the menu items selection has been (optionally) confirmed.
+    //!
+    //! @param b Ignored. It is included in order to match the expected function prototype of the callback method.
+    //
+    function onConfirm(b as Lang.Boolean) as Void {
+        if (mService != null) {
+            mHomeAssistantService.call(mService, {"entity_id"  => mEntity,mData.get("valueLabel").toString() => mValue}, mExit);
```
JosephAbbey: Attributes should be extracted from objects in factory.

source/HomeAssistantNumericMenuItem.mc
```diff
@@ -0,0 +1,182 @@
...
+//
+class HomeAssistantNumericMenuItem extends HomeAssistantMenuItem {
+    private var mHomeAssistantService as HomeAssistantService?;
+    private var mService              as Lang.String?;
+    private var mConfirm              as Lang.Boolean;
+    private var mExit                 as Lang.Boolean;
+    private var mPin                  as Lang.Boolean;
+    private var mData                 as Lang.Dictionary?;
+    private var mValue                as Lang.String?;  
+    private var mFormatString         as Lang.String="%.1f";
```
philipabbey: Only this default value is used. The `mFormatString` is not amended by the constructor. This means an deliberately integer value always has a ".0" on the end.

philipabbey (commented):
Thank you!

philipabbey:
> Rather than using the service/valueLabel way of defining the service call, we define which service to call for entity domains: `light.*`, `input_number.*`, `number.*`, `fan.*`, `valve.*`, `cover.*`, `media_player.*`, `climate.*`.

We might usefully have a `examples\Numeric.md` markdown page to illustrate a range of example numeric items, both float and integer, with the different services calls and value attributes. In this way we could retain a single way of defining to remove potential confusion, whilst spoon feeding users with the currently known domains. @JosephAbbey mentioned the idea of using the schema checker to identify user errors in configuration of known domains.

There's a balance here between:
1. **User configuration error**, mainly because we know how many users don't read the documentation before bleating to us
2. **Keeping the code small & simple** so we don't fall behind changes to Home Assistant (new domains) and add specialised code

The `examples\Numeric.md` markdown page and the schema checks (assuming they are possible) would mitigate the user error to some extent.

source/HomeAssistantNumericMenuItem.mc
```diff
@@ -0,0 +1,231 @@
...
+        if (data == null) {
+            setSubLabel($.Rez.Strings.Empty);
+        } else if(data instanceof Lang.String) {
+            setSubLabel(data);
+        } else if(data instanceof Lang.Number) {
+            var d = data as Lang.Number;
+            setSubLabel(d.format("%d"));
+        } else if(data instanceof Lang.Float) {
+            var f = data as Lang.Float;
+            setSubLabel(f.format(mFormatString));
```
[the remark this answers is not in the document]
JosephAbbey: Actually, I have just realised that `mValue` would be set with the value of the template. What if the user uses inserts units (i.e. `{{ ... }}°C`.

source/picker/HomeAssistantNumericPicker.mc
```diff
@@ -0,0 +1,102 @@
...
+        var val = data.get("start");
+        if (val != null) {
+            start = val.toString().toFloat();
+        } 
+        var step = 1.0;
+        val = data.get("step");
+        if (val != null) {
+            step = val.toString().toFloat();
+        } 
+        val = haItem.getSubLabel().toFloat();
```
philipabbey: I think this is fragile! The template used in the submenu item might contain text. I think @JosephAbbey has suggested a way around this using `attribute` vs `data_attribute`,
see https://github.com/house-of-abbey/GarminHomeAssistant/pull/298#issuecomment-3419695945.

This could be from an API call on opening the Picker, but that would add a delay. So instead include fetching this value in the list of templates evaluated on each fetch of the latest status.

.gitignore
```diff
@@ -7,3 +7,5 @@ Thumbs.db
 source/ClientId.mc
 # Gemini API key for automated translations
 gemini_api_key.txt
+# URLs ans settings modified for testing, so don't sync
+sourc/Settings.mc
```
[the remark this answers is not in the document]
thmichel: Agreed, was for my internal testing anyway and did not work.

source/HomeAssistantNumericMenuItem.mc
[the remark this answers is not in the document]
thmichel: Not necessarily needed as onCOnfirm() will only be called from the Picker which sets mValue first. Could still be set here just to be sage.

JosephAbbey:
@thmichel @philipabbey Can you take a look at the schema proposal: https://github.com/thmichel/GarminHomeAssistant/pull/3

source/HomeAssistantNumericMenuItem.mc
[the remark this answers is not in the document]
JosephAbbey: Yes, the objection was to reading the contents of the label which might not be a number. So I think we should generate a template for value in the same way we do for toggle items with `getToggleTemplate`, that will read the numeric value from the entity and the user can render what ever they want in the subLabel (e.g. add units).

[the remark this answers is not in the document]
philipabbey: Just wondering how much rendering & parsing we are now doing?
- Menu item template for float or int
- Extracting the float or int is now a named attribute instead
- Rendering in the picker, reuse the menu item template? Something without surrounding text?
- API call to set the value, needs to be float or int? Or all forced to float?

thmichel:
> @thmichel @philipabbey Can you take a look at the schema proposal: [thmichel#3](https://github.com/thmichel/GarminHomeAssistant/pull/3)

So, just for my understanding, that ould mean we make min, max, steps mandatory and not optional and also don't use a formatstring any more, not even optional? In that case we would need to define a cofrmat for all different entities like volume, climate etc? Forgive me if I'm dumb here, didn't work with the schema before.

JosephAbbey:
> > @thmichel @philipabbey Can you take a look at the schema proposal: [thmichel#3](https://github.com/thmichel/GarminHomeAssistant/pull/3)
> 
> So, just for my understanding, that ould mean we make min, max, steps mandatory and not optional and also don't use a formatstring any more, not even optional? In that case we would need to define a cofrmat for all different entities like volume, climate etc? Forgive me if I'm dumb here, didn't work with the schema before.

Yes, sorry, let me try and do a better job of explaining myself. We are not going to have smart defaults (at least not yet), so min/max/step/service/attribute/data_attribute will be required fields. My thinking for the format string was that it could be simplified to just a number of decimals to show but i have realised now that you had planned to use it to show units as well? Which does mean that we would either need another property for units, or we go back to having a format string, or we reuse the content value with the rendered template. I think a format string is probably the way to go, because it means we are fully customisable, just needs a good explanation. I will make amendments for this in the schema.

source/HomeAssistantNumericMenuItem.mc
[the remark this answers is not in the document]
thmichel: Do I get it right that with the new json schema, the numeric value of the attribute would be in the data parameter that is being delivered to the updateState function within  the  onReturnUpdateMenuItems function? In that case I could store it in mData to hadle it over to the picker. 
Could someone confirm my thinking is correct?

philipabbey:
> So either we need to put the service attribute back into the tap_action or handle the check for it differently in the HomeAssistantView.

I think @JosephAbbey has been trying to keep our schema in step with HomeAssistant. I'm going to let him comment on this one.

JosephAbbey:
> One thing I've noticed in the new schema: The service attribute is out of the "tap_action" block. Kind of makes sense but would require changes in HomeassistantView.mc as in the initialize function a menuItem is only created when the service attribute exists:
> 
> Line 129:                     
> } else if (type.equals("numeric") && service != null) {
> 
> The service variable is set in line 62: 
>                 if (tap_action != null) {
>                     service = tap_action.get("service");
> 
> So either we need to put the service attribute back into the tap_action or handle the check for it differently in the HomeAssistantView.
> 
> 
> Edit: What works without further modification is the following as  all data is handled over to the numeric menu item in the data field:
> 
>                "tap_action": {
>                     "pin": true,
>                     "confirm": false,
>                     "service": "climate.set_temperature",
>                     "data": {
>                         "step": "0.5",
>                         "display_format": "%.3f",    
>                         "step": "0.5",
>                         "min": "10",
>                         "max": "30"
>                     }
>                }
> Would it make sense to modify the schema accordingly?

There is an argument for putting all of the fields in `tap_action` as they describe what happens when you click, I am not opposed to that if it makes the code simpler/cleaner. But they definitely should not go in the `data` object, that is for data that is sent with the service which is entirely different than data that describes the function of the picker; so it will not be what users expect.

Would you like me to update the schema to have the fields inside the `tap_action`?

source/HomeAssistantNumericMenuItem.mc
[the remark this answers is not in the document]
thmichel: The code is now similar to the toggle menu item, the sublabel is set via the content value  and mValue via the attribute value. This way the user can diplay e.g. the units in the menu item while in the picker only numeric values are shown.

philipabbey:
I've seen you've been working hard. I'm intending to make time to test tomorrow pm. Sorry, I've had domestic commitments so far this weekend.

thmichel:
I've reworked the code so that the numeric menu item can show the content in the sublabel and using attribute to set themValue, similar to the toggle menu item. As this was an initial testing first, I left the values in the data field, like this:

            {
                "type": "numeric",
                "name": "Office",
                "content": "{{ '%.1f °C /  %.1f°C' | format(state_attr( 'climate.xxx' ,'current_temperature'),state_attr('climate.xxx','temperature'))  }}",
                "tap_action": {
                    "pin": false,
                    "confirm": false,
                    "service": "climate.set_temperature",
                    "data": {
                        "display_format": "%.1f",    
                        "step": "0.5",
                        "min": "10",
                        "max": "30",
                        "attribute": "temperature",
                        "data_attribute": "temperature" 
                    }
                },
                "entity": "climate.xxx"
            },

The example displays current and configured temperature with units in the menu item and shows numbers for configured temperature in the picker.
If we don't want to have this in the data field as it is inconsistant with HomeAssistant, I need to look at where in the code is the best place to store it for usage within the numeric menu item. 
As I don't know which option is best (leave it as is, putting the fields into the tap_action or directly at the top level), I leave that entirely up to you and will change code accordingly.

JosephAbbey:
> I've reworked the code so that the numeric menu item can show the content in the sublabel and using attribute to set themValue, similar to the toggle menu item. As this was an initial testing first, I left the values in the data field, like this:
> 
>             {
>                 "type": "numeric",
>                 "name": "Office",
>                 "content": "{{ '%.1f °C /  %.1f°C' | format(state_attr( 'climate.xxx' ,'current_temperature'),state_attr('climate.xxx','temperature'))  }}",
>                 "tap_action": {
>                     "pin": false,
>                     "confirm": false,
>                     "service": "climate.set_temperature",
>                     "data": {
>                         "display_format": "%.1f",    
>                         "step": "0.5",
>                         "min": "10",
>                         "max": "30",
>                         "attribute": "temperature",
>                         "data_attribute": "temperature" 
>                     }
>                 },
>                 "entity": "climate.xxx"
>             },
> 
> The example displays current and configured temperature with units in the menu item and shows numbers for configured temperature in the picker.
> If we don't want to have this in the data field as it is inconsistant with HomeAssistant, I need to look at where in the code is the best place to store it for usage within the numeric menu item. 
> As I don't know which option is best (leave it as is, putting the fields into the tap_action or directly at the top level), I leave that entirely up to you and will change code accordingly.
> 

I can see what you are saying that it simplifies the code to group the options for the picker in an object. How about we do a `picker` object? Which is basically just your `data` object renamed. The only thing that I am really against is putting data that isn't sent to the server in the `data` object.

thmichel:
> I can see what you are saying that it simplifies the code to group the options for the picker in an object. How about we do a `picker` object? Which is basically just your `data` object renamed. The only thing that I am really against is putting data that isn't sent to the server in the `data` object.

That is probably the best option. I'll see through the code and will amend it accordingly. Would you suggest leaving the picker object inside or putting it outside of the tap_action?

source/factory/HomeAssistantNumericFactory.mc
```diff
@@ -0,0 +1,88 @@
...
+
+    //! Generate a Drawable instance for an item
+    //! @param index The item index
+    //! @param selected true if the current item is selected, false otherwise
+    //! @return Drawable for the item
+    public function getDrawable(index as Number, selected as Boolean) as Drawable? {
+        var value = getValue(index);
+        var text = "No item";
+        if (value instanceof Lang.Float) {
+            text = value.format(mFormatString);
```
philipabbey: Monkey C is a pain!

There seems to be no way to include a `%` sign in the format string without a crash. So if you wanted to format 0%..100% in the number picker, I don't yet see how you can.

See [Float.format](https://developer.garmin.com/connect-iq/api-docs/Toybox/Lang/Float.html#format-instance_function) docs.

I tried:
```
value.format("%d%%")
value.format("%d%")
value.format("%d\%")
```

source/HomeAssistantApp.mc
```diff
@@ -629,6 +629,10 @@ class HomeAssistantApp extends Application.AppBase {
                             if (item instanceof HomeAssistantToggleMenuItem) {
                                 (item as HomeAssistantToggleMenuItem).updateToggleState(data[i.toString() + "t"]);
                             }
+                            if (item instanceof HomeAssistantNumericMenuItem) {
+                               // (item as HomeAssistantNumericMenuItem).updateNumericState("22");
+                                (item as HomeAssistantNumericMenuItem).updateNumericState(data[i.toString() + "n"].toString());
```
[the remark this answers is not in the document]
philipabbey: This was probably because I used an integer number with a `"%.1f"` format string. I think this indicates how easy it will be for users to misconfigure (and moan at us).

source/factory/HomeAssistantNumericFactory.mc
philipabbey: With these two review comments, I am wondering if we need to be more safe and less ambitious? An alternative to the format string is to specify an integer number of decimal places from 0..n. We might also be able to extract that from the `step` size. E.g. given `step`

1 => `"%d"`
0.1 => `"%.1f"`
0.02 => `"%.2f"`

etc. This could be encoded in a function that identifies the minimum multiples of 10 required to make the step size >= 1.

The downside is we lose the ability to format the number in the picker. But its one thing less to configure? Its a pity, but I'm worried that HA users will not all be familiar with C's `printf(..)` syntax, let alone the Monkey C variation.

philipabbey:
@thmichel I'm slightly confused by the reply. Were you in favour of a default formatter of some kind or in favour of the function to create the formatter from the steps? There were a few negatives in quick succession and I lost the sense of your reply.

philipabbey:
> I would vote for option two as it's more flexible if new devices or scripts are added.

I've implemented option 2, and its working well enough. There's some oddness going on with the `attribute` field, which is not getting through to the menu item. Really odd.

source/factory/HomeAssistantNumericFactory.mc
```diff
@@ -44,11 +44,15 @@ class HomeAssistantNumericFactory extends WatchUi.PickerFactory {
...
-
+        if (mStep < 0.01) {
+            mFormatString="%.3f";
+        } else if (mStep < 0.1) {
+            mFormatString="%2f";
+        } else if (mStep < 1) {
+            mFormatString="%.1f";
+        } else {
+            mFormatString="%d";
+        }
```
philipabbey: I have a more general solution.

philipabbey:
> Edit: Didn't see your previous comment. Not sure what you mean with "retracted"? Sorry, I'm new to all this git collaboration stuff

I mentioned that I was willing to sort some of the changes to the code. You suggested that would be fine. I'm now doing that. As a result I branched from your PR two commits ago. That means there will be merge conflicts.

Really sorry about this, but I will try and merge you code into my PR on your PR. Then in doing so I will request my changes over yours and negate your work in favour of mine. I just need a bit more time. I was suggesting you hold fire until I can get you my PR to review.

source/factory/HomeAssistantNumericFactory.mc
philipabbey: See https://github.com/thmichel/GarminHomeAssistant/pull/4.

PR #301, opened 2025-10-22
Add light effect selector example to Select.md

josephabbey:
Added an example of a light effect selector in JSON format.


JosephAbbey:
Needs amended nav bar

philipabbey:
Like it, nice example.

How is this linked into the wider documenation?
Do we need a menu at the head of the page? In which case we need to amend the headers in `*.md` too.
Or should we add this to `examples/Actions.md`?

PR #306, opened 2025-10-30
305 code tidy documentation for v36

philipabbey:
1. Amended `Lang.Dictionary` handling so more of them use the `dict["key"] = value` format.
2. Added documentation for the `numeric` menu type.

### DIFF
--- a/examples/Numeric.md
+++ b/examples/Numeric.md
@@ -112,6 +112,30 @@ For copy and paste, the Jinja2 fields are as follows:
 {{ not is_state('media_player.amplifier','unavailable') }}
 ```
 
+As an alternative to using the GUI, the following can be pasted into Home Assistant's `configuration.yaml`:
+
+```yaml
+template:
+  - number:
+      - name: "Amplifier dB"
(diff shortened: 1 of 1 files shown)


examples/Numeric.md
```diff
@@ -0,0 +1,139 @@
...
+```
+{{ (value+80)/100 }}
+```
+
+3. Availability template:
+
+```
+{{ not is_state('media_player.amplifier','unavailable') }}
+```
+
```
JosephAbbey: There is a way to define this with YAML, we should probably show how to do that as well. https://www.home-assistant.io/integrations/template/#number

examples/Numeric.md
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+},
+```
+
+## Trouble Shooting
+
+Specific to this menu item:
+
+1. If the number picker does not initialise with the correct value, amend the `attribute` field. Just because your template renders does not mean the application has extracted the numeric valueas the `content` template is rendered on the Home Assistant server.
```
JosephAbbey: `valueas` typo

PR #346, opened 2026-03-03
342 allow two user supplied http headers

philipabbey:
This should enable enough headers for username and password type situations.

Please don't merge yet as I need to re-verify on watch after the last pull request was merged to `main` and `main` merged to here.


josephabbey (approved):
Took a long time to scroll through those translation changes.

PR #350, opened 2026-03-10
Send last known location in Background Service

### LINKED ISSUE #348: Send GPS location with last-known quality
As of now, the background service is only sending GPS location update when the Quality is neither not available nor last known. This implies that for the app to send a position update, the device needs to be currently getting a GPS fix so this basically means that the device is actively in an activity and not simply worn by the user. I am suggesting allowing Position.QUALITY_LAST_KNOWN to determine if we should send a position update. The garmin doc states "The Location is based on the last known GPS fix." when the Quality is QUALITY_LAST_KNOWN. Making this change would allow the app so send the last known location of the device meaning that an updated location is not only sent when it happens that the GPS fix is done at the same moment the background service task is called. It would allow easier automation based on the location of the device in HA.

If we want to keep the precision of the current location, we could simply add a new entity with last known location which could be used and the user would simply assume a lower accuracy of that position compared to the other location entity.

philipabbey:
Changes to location sending and accuracy.

### DIFF
--- a/source/BackgroundServiceDelegate.mc
+++ b/source/BackgroundServiceDelegate.mc
@@ -115,15 +115,17 @@ class BackgroundServiceDelegate extends System.ServiceDelegate {
         // System.println("BackgroundServiceDelegate onTemporalEvent(): Activity : " + Activity.getProfileInfo().name);
 
         // Don't use Settings.* here as the object lasts < 30 secs and is recreated each time the background service is run
-
-        if (position.accuracy != Position.QUALITY_NOT_AVAILABLE && position.accuracy != Position.QUALITY_LAST_KNOWN) {
+        if (position.accuracy != Position.QUALITY_NOT_AVAILABLE) {
             var accuracy = 0;
             switch (position.accuracy) {
+                case Position.QUALITY_LAST_KNOWN:
+                    accuracy = 100;
+                    break;
                 case Position.QUALITY_POOR:
-                    accuracy = 500;
+                    accuracy = 100;
                     break;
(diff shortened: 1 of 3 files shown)


source/BackgroundServiceDelegate.mc
```diff
@@ -115,15 +115,17 @@ class BackgroundServiceDelegate extends System.ServiceDelegate {
         // System.println("BackgroundServiceDelegate onTemporalEvent(): Activity : " + Activity.getProfileInfo().name);
 
         // Don't use Settings.* here as the object lasts < 30 secs and is recreated each time the background service is run
-
-        if (position.accuracy != Position.QUALITY_NOT_AVAILABLE && position.accuracy != Position.QUALITY_LAST_KNOWN) {
+        if (position.accuracy != Position.QUALITY_NOT_AVAILABLE) {
             var accuracy = 0;
             switch (position.accuracy) {
+                case Position.QUALITY_LAST_KNOWN:
+                    accuracy = 100;
```
josephabbey: I think it would be better to go for at least 200 here, so that in a homeassistant automation you can read this attribute to detect if a location is last known.

josephabbey (changes requested):

source/BackgroundServiceDelegate.mc
philipabbey: Do you want to post that on the issue #348 where we had the discussion?

philipabbey: Amended, merge if happy.

josephabbey (approved):

PR #362, opened 2026-06-22
Preserve tap action data for numeric picker actions

StumblingGamer:
Numeric menu items currently discard existing `tap_action.data` when confirming a picker value. This prevents service call options such as `transition` from being sent with `light.turn_on`.

This change passes `tap_action.data` into numeric menu items and merges that data with the selected numeric picker value before calling the Home Assistant service.

Example fixed payload:

```json
{
  "entity_id": "light.living_room",
  "transition": 2,
  "brightness": 128
}

Tested with a Forerunner 965 build and a Home Assistant light brightness numeric picker.

### DIFF
--- a/config.schema.json
+++ b/config.schema.json
@@ -256,6 +256,11 @@
                 "action": {
                   "$ref": "#/$defs/action"
                 },
+                "data": {
+                  "type": "object",
+                  "title": "Additional action data",
+                  "description": "Fixed Home Assistant service data. The selected picker value is merged into this object using the picker's data_attribute. Service-specific fields cannot be schema checked here."
+                },
                 "picker": {
                   "type": "object",
                   "title": "Number picker configuration",
--- a/examples/Numeric.md
+++ b/examples/Numeric.md
@@ -39,6 +39,36 @@ Field            | Purpose                                                      
 
 It may well be the case that often `attribute` and `data_attribute` are the same attribute, as with this example.
 
+## Additional Action Data
+
+The `tap_action` object may also include a `data` object for fixed service call parameters. When the picker value is selected, the application merges the picker value into this `data` object using `data_attribute`.
+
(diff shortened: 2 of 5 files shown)


philipabbey:
Hi @StumblingGamer thank you for offering this code and being prepared to work with the Garmin SDK to produce it.

Please can I ask that you include or propose the changes required to the follow too?:
1. documentations at https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/examples/Numeric.md
2. JSON schema at https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/config.schema.json#L253. I think it might just need a field like this one at https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/config.schema.json#L953 defined. Presently your changes should be failing schema checking (I hope ;-). We need the checker detailed at https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/TroubleShooting.md#check-your-json-schema to pass with your amendments for the `tap_action.data` field.

Thank you!

Philip

StumblingGamer:
Hello @philipabbey, I have made the requested updates.  

I tested the numeric example with tap_action.data in the documented web editor. The current schema did not appear to reject the field, likely because numeric tap_action does not currently forbid additional properties at that nested level. I still added data explicitly to the numeric schema so it is documented, discoverable, and type-checked as an object.

Thanks for your consideration!
Dave

philipabbey:
Excellent! It all looks tidy and we have identified a local test (important in case we have to support the code in the future). I hope to do this by the end of the weekend.

> The current schema did not appear to reject the field, likely because numeric tap_action does not currently forbid additional properties at that nested level.

Noted

Thank you!

StumblingGamer:
Thanks!

philipabbey (approved):
Working on our own tests.

philipabbey:
Version 3.14 now released. Thank you again.

PR #363, opened 2026-06-24
361 prevent pin items for non touch screens

### LINKED ISSUE #361: Prevent PIN items for non touch screens?
Is there any sensible way to manage non-touch screens trying to use the PIN view?

A touch screen enabled device can be determined by [`System.DeviceSettings.isTouchScreen`](https://developer.garmin.com/connect-iq/api-docs/Toybox/System/DeviceSettings.html#isTouchScreen-var). Perhaps refuse to render any menu item with PIN confirmation requested when that result is false? Rendering the menu item without using the PIN would be a security risk by definition.

philipabbey:
1. Originally intended bug fix:

image

2. Fixes for the schema regarding numeric `tap_action`.
3. Code fix to initialise numeric pickers when no user defined template is supplied.
4. Amended version history.


josephabbey (approved):

josephabbey:
Looks good!

PR #374, opened 2026-08-21
Documentation update

### LINKED ISSUE #292: Unable use over WiFi/LTE
I see the setting enable execution commands over WiFi/LTE. I’m assuming I can use the app without my phone around & trigger via WiFi. However when I do this I get an error saying phone is not connected. I see on known bugs that you cannot trigger a toggle if it is showing the incorrect status, but it does the same thing on a tap as well.

philipabbey:
@xhemart Please can you cast your eye over the revised documentation and suggest any amendments necessary?


xhemart:
Thanks for pulling this together and for the credit! A couple of things before I approve:

**Typos in Wi-Fi.md:** "TSL settings" → "TLS settings", and "its appears" → "it appears".

**Possible inconsistency between TroubleShooting.md and Wi-Fi.md:** the new "Wi-Fi Not Working" section in TroubleShooting.md still frames this as "a TLS issue on your Home Assistant server" that's "specific to your IT setup" — which was the pre-diagnosis explanation. The corrected explanation a few lines down in Wi-Fi.md says the opposite: it's a device-side cipher-suite limitation (no forward secrecy offered over Wi-Fi), not a server misconfiguration, and there's no real app- or server-side fix short of degrading the server's security by allowing legacy RSA key exchange. Might be worth either updating TroubleShooting.md to match, or trimming it down to just point at Wi-Fi.md rather than restating the older theory.

**One nuance on the model list:** grouping "Epix (Gen 2), Fenix 7 Pro, Forerunners 265 & 970" together might overstate the evidence for the 970 — that report was intermittent and self-resolved after a reset without a confirmed cause, and the simulator reproduction of "No Internet connection" isn't necessarily the same TLS/cipher-suite failure (could be an unrelated simulator quirk). The other three have more direct evidence (reproducible error code 0, or the cipher-suite test itself). Might be worth a lighter caveat there, e.g. "confirmed on Epix (Gen 2); likely related on Fenix 7 Pro and Forerunner 265; possibly related on Forerunner 970 (unconfirmed)."

Otherwise this is a great writeup — thanks for digging into it.

philipabbey:
Yes, I had created the new trouble shooting section prior to your more precise diagnosis.

> This is probably a TLS issue on your Home Assistant server.

Changed to "This is probably a TLS issue associated with HTTPS."

Thank you for your attention to the details, especially the evidence for the devices.

philipabbey:
@xhemart I'm keen to get this merged as there's a new PR come in.

xhemart (approved):
Approving to unblock the merge — the typo and model-list points are fully addressed. Happy to leave the remaining TroubleShooting.md/Wi-Fi.md wording nuance as-is; not a blocker. Thanks for turning this around quickly!

xhemart:
Quick, slightly embarrassing follow-up on the cipher-suite explanation in this PR (and in my earlier comment on #292) — sorry to revise this again, especially now that it's already merged, but I think it's worth flagging.

I ended up doing a packet capture of my Epix (Gen 2)'s actual TLS traffic at home (isolated it from my phone, confirmed the source MAC belongs to Garmin, reproduced it twice). Turns out my watch's ClientHello does offer ECDHE cipher suites — contrary to what I assumed from the Garmin forum bug report — and I verified directly against my own Nabu Casa endpoint that this exact cipher list negotiates successfully.

So at least for my unit/firmware, "the watch doesn't support ECDHE" doesn't seem to hold up. I don't have a confirmed replacement cause yet — my current best guess is something around certificate chain validation (Nabu Casa's cert comes from Let's Encrypt's newest CA hierarchy, rolled out about a year ago, which an older/embedded trust store might not recognize yet), but that's genuinely unverified on my end, so I don't want to swap one unverified claim for another in the docs.

Doesn't change anything practical — Bluetooth still works fine, and there's no app-side fix either way — but wanted to flag it in case the docs are worth a small follow-up edit at some point. Happy to help with that if useful. Sorry for the churn on this one!

philipabbey:
Thanks for being up front, and I'm not worried. If you want any of the text retracted in the meantime feel free to put a PR in. I am reliant on the better knowledge of others here, a benefit of open source. I'll be able to approve the PR pretty much instantly as I'm learning from this technical exchange!

Issue #347, opened 2026-03-08
Garmin Instinct 3 Solar 45mm - Glance not working


toto99303:
Hello,

I have Garmin Instinct 3 Solar 45mm and everything works when I start the app from applications & activities. I can switch lights on and off, all good. However when I put the app in Glance favourites it stays only as name and does the same as in applications & activities, eg. you can start the app and toggle whatever is in the menu.json.
I thought I should be seeing API status or Custom status if I set one?
I added the glance section in my config, but it's still not visible as widget.
Is this intentional on the black and white screens, or is there a problem on my side?
Thanks in advance!

![Image](https://github.com/user-attachments/assets/196a9948-8f31-4321-97bd-cb21ad2fa372)

philipabbey:
> I thought I should be seeing API status or Custom status if I set one?

Yes, from https://developer.garmin.com/connect-iq/device-reference/instinct3solar45mm/ I believe your device does support glances.

The Glance will only work is you have enabled caching, see the explanation at https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/examples/Glance.md

Also, which version have you installed?
- [GarminHomeAssistant (Widget)](https://apps-developer.garmin.com/apps/559f5174-177f-4f46-b170-f31c7e74dea3)
- [GarminHomeAssistant (Application)](https://apps-developer.garmin.com/apps/61c91d28-ec5e-438d-9f83-39e9f45b199d)

You will need the second one for glances to work. (The widget version is provided for some older devices and those who preferred it.)

Complete aside:
- From #132, we know the menu icons don't work and its too awkward to fix.

toto99303:
Thanks for replying. Yes, I have enabled menu cache and I am using the application version of the app. Still, it appears as above. I tried everything, but I can't make it work.

Watch software version is 13.29

philipabbey:
Sorry for the basic questions. Do any of your other glances provide dynamic data? Not possible to tell from your photo. Just wondering if glances do actually work on your model?

toto99303:
> Sorry for the basic questions. Do any of your other glances provide dynamic data? Not possible to tell from your photo. Just wondering if glances do actually work on your model?

Yes, they do, everything else is normal.

![20260309_073727.jpg](https://github.com/user-attachments/assets/ca3f9cdb-a715-4437-afcb-5036d2383cc2)

philipabbey:
[...]

PR #375, opened 2026-08-23
feat: add select menu item type

abstractionnl:
Add a new 'select' menu item type that presents a scroll-wheel picker for choosing from a list of string options, then calls a Home Assistant service with the selected value.

Features:
- Entity-based mode: options auto-fetched from entity's 'options' attribute
- Manual mode: options specified as value/label pairs or plain strings
- Action defaults to select.select_option or input_select.select_option based on entity domain, but any entity/action combination is supported
- data_attribute defaults to 'option'
- Supports confirm, pin, and exit flags
- Current selection shown as sub-label, refreshed via template polling

New files:
- HomeAssistantSelectMenuItem.mc - menu item class
- HomeAssistantSelectFactory.mc - picker factory for string options
- HomeAssistantSelectPicker.mc - picker view and delegate
- examples/Select.md - usage documentation
- select_type.svg icons for all 16 resolution folders

Modified files:
- config.schema.json - added select type definition and examples
- HomeAssistantView.mc - parsing, getItemsToUpdate, onSelect handler
- HomeAssistantMenuItemFactory.mc - select factory method with icon
- HomeAssistantApp.mc - template keys 's' and 'o' for state polling
- All drawables.xml and navigation bars updated


philipabbey:
Sounds like a significant piece of work. Just to acknowledge your pull request, and promise to take a look.

philipabbey:
Reviewing this with interest. I have deployed a test app and got the select functionality working from my watch, successfully following your instructions. You mention support for confirmation and PIN, you do also support `enable` too, which might be worth adding to the docs.

philipabbey:
Your `Select.md` file had best reference the existing `tap_action` alternative. It's just a cycle through the values, and not as fully functional as your new code, but does provide an alternative that might suit some users. See https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/examples/Actions.md#selects. I've just updated that documentation on this PR if you don't mind as the examples are related if contrasting.

source/HomeAssistantSelectMenuItem.mc
```diff
@@ -0,0 +1,241 @@
...
+        if (mAction != null && mSelectedValue != null) {
+            var data = {} as Lang.Dictionary;
+            if (mData != null) {
+                var keys = mData.keys();
+                for (var i = 0; i < keys.size(); i++) {
+                    data[keys[i]] = mData[keys[i]];
+                }
+            }
+            data["entity_id"] = entity_id.toString();
+            data[mDataAttribute] = mSelectedValue;
```
josephabbey: I think the selected value should really be passed to this function as a parameter like the ToggleMenuItem, it prevents any possible race condition where the mSelectedValue gets updated back to its current value before the command is sent.

josephabbey (changes requested):
Just one small thing, otherwise I really like this! Thank you!

philipabbey:
My next check is for memory usage given we're on the limit with Instict devices presently. See https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/Devices.md. If we break Instinct devices by adding more code I know I'll have an unhappy user, so this could be a deciding factor.

Also you commit https://github.com/house-of-abbey/GarminHomeAssistant/pull/375/commits/41907fa308f7d1e07aa055902107a1d3381a6555 worked nicely for me as I have only used the minimal JSON in your first example so far.

josephabbey:
@philipabbey I have noticed while reviewing this that the `HomeAssistantToggleMenuItem.updateState` function uses a bear `Communications.makeWebRequest`, should it be using the `HomeAssistantService`?

philipabbey:
> @philipabbey I have noticed while reviewing this that the `HomeAssistantToggleMenuItem.updateState` function uses a bear `Communications.makeWebRequest`, should it be using the `HomeAssistantService`?

`HomeAssistantToggleMenuItem.setState`, not `updateState` nor `updateToggleState`. I think that's where the menu items diverged such that we could not share the `HomeAssistantService`.

philipabbey:
How annoying! Compiling for the `instinct2x` device in order to check memory usage as per previous post.

```text
ERROR: instinct2x: A:\Philip\Programming\MonkeyC\HomeAssistant\source\HomeAssistantSelectMenuItem.mc:34: Too many arguments passed to method 'initialize'. Only 9 arguments are allowed.
```

There are 10 parameters here.

```text
    function initialize(
        label            as Lang.String or Lang.Symbol,
        template         as Lang.String,
        action           as Lang.String?,
        data             as Lang.Dictionary?,
        dataAttribute    as Lang.String,
        labels           as Lang.Array<Lang.String>,
        values           as Lang.Array<Lang.String>,
        hasManualOptions as Lang.Boolean,
        options          as {
            :alignment as WatchUi.MenuItem.Alignment,
            :icon      as Graphics.BitmapType or WatchUi.Drawable or Lang.Symbol,
            :exit      as Lang.Boolean,
            :confirm   as Lang.Boolean,
            :pin       as Lang.Boolean
        }?,
        haService        as HomeAssistantService
    ) {..}
```

Argh! Compiles fine on `venu2`. Is there anything we can amend here?

philipabbey:
Just reviewed the code. That's a serious amount of work done. Thank you for your effort.

abstractionnl:
> How annoying! Compiling for the `instinct2x` device in order to check memory usage as per previous post.
> 
> ```
> ERROR: instinct2x: A:\Philip\Programming\MonkeyC\HomeAssistant\source\HomeAssistantSelectMenuItem.mc:34: Too many arguments passed to method 'initialize'. Only 9 arguments are allowed.
> ```
> 
> There are 10 parameters here.
> 
> ```
>     function initialize(
>         label            as Lang.String or Lang.Symbol,
>         template         as Lang.String,
>         action           as Lang.String?,
>         data             as Lang.Dictionary?,
>         dataAttribute    as Lang.String,
>         labels           as Lang.Array<Lang.String>,
>         values           as Lang.Array<Lang.String>,
>         hasManualOptions as Lang.Boolean,
>         options          as {
>             :alignment as WatchUi.MenuItem.Alignment,
>             :icon      as Graphics.BitmapType or WatchUi.Drawable or Lang.Symbol,
>             :exit      as Lang.Boolean,
>             :confirm   as Lang.Boolean,
>             :pin       as Lang.Boolean
>         }?,
>         haService        as HomeAssistantService
>     ) {..}
> ```
> 
> Argh! Compiles fine on `venu2`. Is there anything we can amend here?

I think maybe the hasManualOptions can be removed, it is a derived value. I will look into it.

abstractionnl:
> My next check is for memory usage given we're on the limit with Instict devices presently. See https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/Devices.md. If we break Instinct devices by adding more code I know I'll have an unhappy user, so this could be a deciding factor.
> 
> Also you commit [41907fa](https://github.com/house-of-abbey/GarminHomeAssistant/commit/41907fa308f7d1e07aa055902107a1d3381a6555) worked nicely for me as I have only used the minimal JSON in your first example so far.

I see the instinct2 is very memory constrained. I have been testing on vivoactive6 and venu3, which have "plenty" of memory. I  hoped adding a little amount of code would still fit, even if it cannot be used on older devices.

abstractionnl:
> Just reviewed the code. That's a serious amount of work done. Thank you for your effort.

Not a problem, thank you for the review. I am completely new to programming for Garmin devices so please look at it carefully. I will try to address the remaining comments soon.

I tested this feature for a few days and I really like the addition of quickly being able to have a menu item dedicated to selects, instead of having to resort to a group with numerous tap items in it. So I hope this feature can make it into the codebase.

philipabbey:
@abstractionnl Were there any more commits ready?

abstractionnl:
> @abstractionnl Were there any more commits ready?

I think it must be okay now. I managed to use the HomeAssistantService and fix the 10 parameter constructor. Not sure about the memory constraint for instinct2x. I cannot run the emulator with my configuration, but that is unrelated to this change as it is just too big for that device, venu2 is fine. Let me know what you think.

philipabbey:
Working on larger or new devices. Broken on `instinct2x` , related issue is #347.

@toto99303 this is a rather nice addition to the application, only I fear it means the end of support for devices with 98k application memory. The table at https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/Devices.md lists 5 devices that would no longer be able to keep up. Are you still using your Instinct device?

toto99303:
> Working on larger or new devices. Broken on `instinct2x` , related issue is #347.
> 
> @toto99303 this is a rather nice addition to the application, only I fear it means the end of support for devices with 98k application memory. The table at https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/Devices.md lists 5 devices that would no longer be able to keep up. Are you still using your Instinct device?

Hello, yes, I am still using it daily with my HA :-)

philipabbey:
More investigation required, but this PR seems to be demanding watches with 524,288 bytes of memory or more. Thats an FR255 at minimum. That wipes out 19 devices currently in support.

That's awkward...

abstractionnl:
> More investigation required, but this PR seems to be demanding watches with 524,288 bytes of memory or more. Thats an FR255 at minimum. That wipes out 19 devices currently in support.
> 
> That's awkward...

It feels quite weird that the addition of a small part of code results in a much higher memory usage.
For reference, I have quite an elaborate profile, and simulation on venu2 take about 100kb, 
The simulation seems to work for fr245 or vivoactive3 (with 128k).

philipabbey:
I was struggling with `vivoactive3`, but wondered if it was inclusion of debug. I agree its an unexpected increase. I'm finding a large number of devices simply not simulating with the latest SDK. So I presently doubt my findings.

philipabbey:
So after significant effort finding a way to compile the application withotu debug and provide some basic settings temporarily so the application will run I now have the following for the `instinct2x` device:

```text
Error: Out Of Memory Error
Details: 'Failed invoking <symbol>'
Time: 2026-09-04T11:24:38Z
Part-Number: 006-B4394-00
Firmware-Version: '14.12'
Language-Code: eng
ConnectIQ-API-Version: 3.4.3
ConnectIQ-Version: 6.0.2
Filename: HOMEASSISTANT
Appname: HomeAssistant
Stack:
  - pc: 0x100022fb
    File: 'A:\Philip\Programming\MonkeyC\HomeAssistant\source\HomeAssistantApp.mc'
    Line: 277
    Function: fetchMenuConfig
  - pc: 0x10001cdf
    File: 'A:\Philip\Programming\MonkeyC\HomeAssistant\source\HomeAssistantApp.mc'
    Line: 142
    Function: getInitialView
```

`vivoactive3` is running perfectly. Given how tight the remaining memory was after reading the menu definition on Instinct devices, I'm not surprised. Inspite of the code being efficiently written, this PR does add significant new code.
- `source/HomeAssistantSelectMenuItem.mc`
- `source/HomeAssistantSelectFactory.mc`
- `source/HomeAssistantSelectPicker.mc`

The affected devices will be:
- `descentg1`
- `instinct2`
- `instinct2s`
- `instinct2x`
- `instinctcrossover`

This means we have a decision to make:
1. Decline the PR after much useful and hard work.
2. Drop 5 devices from support, meaning they should never be updated and they'll have to stick to Version 3.15 for ever.
3. Fork the code. We did that for the [Widget version](https://github.com/house-of-abbey/GarminHomeAssistantWidget) and it was maintained for a short while and then allowed to slowly rot.

It feels like (2) is the way forwards as it should allow Instinct users to continue using the application, just without updates until they change their watches. @toto99303 Can you live with that?

Aside:

> > @philipabbey I have noticed while reviewing this that the `HomeAssistantToggleMenuItem.updateState` function uses a bear `Communications.makeWebRequest`, should it be using the `HomeAssistantService`?
> 
> `HomeAssistantToggleMenuItem.setState`, not `updateState` nor `updateToggleState`. I think that's where the menu items diverged such that we could not share the `HomeAssistantService`.

Ah, because the method concerned was only used in one class and there was no immediate need to factor it out into the service class. But it might be more elegant.

abstractionnl:
Is the package we deliver a single app, or can we add compiler directives to differentiate between them?

philipabbey:
> Is the package we deliver a single app?

Its one app per device. Adding more devices seems to make the compilation (export) product ever larger.

> can we add compiler directives to differentiate between them?

There does seem to be some mechanism based on these pages, or at least there was once:
- https://forums.garmin.com/developer/connect-iq/f/discussion/4298/conditional-compilation-in-monkeyc seems to suggest you can use a device name for condition compilation.
- https://forums.garmin.com/developer/connect-iq/f/discussion/6376/how-to-exclude-files-and-folders-with-jungles

Now this is quite new and possibly worthwhile exploring if the support still exists. The latter suggest using annotations, and they do work, so we might need something like this in `monkey.jungle`:

```text
instinct2x.excludeAnnotations=selectView
```

Then use `(:selectView)` to label the parts to omit. Other tricks might be required where the select view code meets the common code, e.g. using `has` to test for the inclusion of code for a class. E.g.

```
if ((Activity has :getActivityInfo) and (Activity has :getProfileInfo)) { .. }
```

The latter part will still compile some additional code into the `instinct2x` version of the application, so we need to keep that small.

Hmm, I'm curious to try this out.

philipabbey:
I've got a local version to compile and run on an `instinct2x` device, but it crashes with:

```text
HomeAssistantApp onReturnFetchMenuConfig() Memory: total=94112, used=80864, free=13248

Error: Out Of Memory Error
Details: Failed invoking <symbol>
Stack: 
  - onReturnFetchMenuConfig() at A:\Philip\Programming\MonkeyC\HomeAssistant\source\HomeAssistantApp.mc:222 0x10003059 

Encountered app crash.
```

This means the app ran long enough to fetch the menu definition file and then ran out of memory to build the menu. Link to the crash point is https://github.com/house-of-abbey/GarminHomeAssistant/blob/main/source/HomeAssistantApp.mc#L222. Its at the point where it tries to save the fetched menu to `Storage`.

I could commit the code to this PR, but its known to not work. The code is littered with `($ has :HomeAssistantSelect<blah>)` tests, and then I had to test individual methods' presence too when they were called too. Quite messy.

Still works on a `venu2` device with full functionality.

Hang-on, my batch file for no debug failed to include the necessary `--release` switch for no debug. Checking back over the steps, I can get the application to run on an `instinct2x` device with a decent sized menu if:
1. Use annotations to omit the new code
2. No Debug

I'll push my solution, but it needs a code review of my changes in case there's a better way to achieve the result with less vestigial code.

See `compile_sim.cmd` for an example of how to run without debug, outside of VS Code.

philipabbey:
I've got a second potential user asking me to release the app for the new Fenix 9 device released on 25 Aug 2026. I mean _"why is the app not already compiled for the device?"_ Clearly I'm slacking. So is there any chance we can finish this PR up? Then I can release this and the inclusion of Fenix 9 in the next release.

Are there any improvements you would like to offer on my code changes?

Thanks!

philipabbey:
@abstractionnl Please excuse my forced push, it corrects a serious mistake.

josephabbey (approved):
Looking good!

philipabbey (approved):

abstractionnl:
Looking good. Going to test tomorrow.

source/HomeAssistantView.mc
```diff
@@ -182,19 +182,21 @@ class HomeAssistantView extends WatchUi.Menu2 {
...
-                                data,
-                                tap_action,
-                                {
-                                    :exit    => exit,
-                                    :confirm => confirm,
-                                    :pin     => pin
-                                }
-                            ));
+                        } else if (($ has :HomeAssistantSelectMenuItem) && type.equals("select") && (entity != null)) {
+                            if (HomeAssistantMenuItemFactory has :select) {
```
abstractionnl, line 186: Can't we assume that since HomeAssistantSelectMenuItem exists, this must also exist?

philipabbey: I don't think so. Try removing it and compiling?

The test for existence of `HomeAssistantSelectMenuItem` does not cover the existence test for a method `select` in a different class `HomeAssistantMenuItemFactory`. Note you can conditionally compile both classes and methods separately. So in `HomeAssistantMenuItemFactory` I have a conditional method.

```text
    (:selectView)
    function select(..)
```

I only added sufficient conditions to achieve compilation. That means each test I added solved a compilation problem. The real question is, is there another structure to the code that reduces all those additional tests?

abstractionnl: I see, it does not work, even though the code could not be called anyway.

abstractionnl:
Changes LGTM.

philipabbey:
Thank you for your contribution.

philipabbey:
@abstractionnl & @toto99303 Version 3.16 now uploaded, please let me know all it okay. In particular that we've not broken Instinct devices! Thank you.

